PR #1, opened 2016-07-27
Create an Event Map for EIDR Connect

mawaruDOLL:
I repurposed the event map code so that it works with tracked events.  I added a map to the event details page as well as plotted all locations on the map under the Event Map menu option. The event map currently doesn't handle plotting multiple events at the same location, so you'll only see one of the events. We may need a clustering plugin for leaflet to take care of that.

### DIFF
--- a/client/controllers/eventMap.coffee
+++ b/client/controllers/eventMap.coffee
@@ -32,18 +32,20 @@ Template.eventMap.rendered = ->
     map.removeLayer(markers)
     markers = new L.FeatureGroup()
     query = instance.query.get()
+    
     if _.isObject query
       filteredEvents = instance.data.events.find(query).fetch()
     else
       map.removeLayer(markers)
       return
     for event in filteredEvents
-      if event.locations
-        name = event.eventNameVal
-        eidID = event.eidID
+      eventLocations = instance.data.locations.find({userEventId: event._id}).fetch()
+      if eventLocations.length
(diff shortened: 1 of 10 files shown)


nathanathan:
I'm seeing a bug in the event map. When I added the location "Mount Seattle" to an event, then looked at the event map, the marker was in the wrong place, and when I zoomed in on it, the marker started moving around. I saw the following error in my browser console:
```
TypeError: Cannot read property 'lat' of null
    at Object.project (fuatsengul_leaflet.js:14)
    at Object.latLngToPoint (fuatsengul_leaflet.js:14)
    at e.project (fuatsengul_leaflet.js:14)
    at e.latLngToLayerPoint (fuatsengul_leaflet.js:14)
    at e.update (fuatsengul_leaflet.js:15)
    at e.onAdd (fuatsengul_leaflet.js:15)
    at e._layerAdd (fuatsengul_leaflet.js:14)
    at e.addLayer (fuatsengul_leaflet.js:14)
    at eachLayer (fuatsengul_leaflet.js:15)
    at e.onAdd (fuatsengul_leaflet.js:15)
fuatsengul_leaflet.js:14 Uncaught TypeError: Cannot read property 'lat' of null
```

nathanathan:
Never-mind, it turns out the bug I was seeing was caused by old location data.

Nice work!

PR #2, opened 2016-07-27
Clean up forked version of EIDR

mawaruDOLL:
I removed the code dealing with historical events and commented out some menu options that don't work right now but may be used in the future.


nathanathan:
👍

PR #4, opened 2016-07-28
When EIDR C user not logged in, the articles area has a log in form

mawaruDOLL:
I removed the log in form from the articles section and fixed up some of the styling.

### DIFF
--- a/client/views/userEvent.jade
+++ b/client/views/userEvent.jade
@@ -77,8 +77,8 @@ template(name="articles")
     .panel-heading
       h3 Articles
     .panel-body.comments-container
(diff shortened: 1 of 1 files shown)


mawaruDOLL:
I realized the spacing was off in the articles template. I fixed that and removed some css rules I added since they ended up not being all that necessary. I added a rule in a different branch that will deal with the padding of a p tag in a panel, which is used when no articles are associated with the event.

sturmer-eha:
Looks good. I'd squash those two commits into one since it was a fix in the latter one.
👍

PR #5, opened 2016-07-28
Create audit trail for event locations

mawaruDOLL:
I restyled the location section so the layout matches the articles section. Now each location link expands a details section that contains the geonames link, the user who added it, and the date it was added. The delete button still looks weird on extra small view ports though. But I didn't want to make it fill all 12 columns because that looked even more cluttered.

I keep using styling from the comments stylesheet for these sections. I think at some point it should be renamed/rewritten to describe event sections without the naming being specific to comments.


sturmer-eha:
👍

PR #6, opened 2016-07-29
Users can select article publication dates

mawaruDOLL:
I added a field to the article form for including the publication date of the article.  While I was in that area of the code, I also included some auditing information and updated the event's last modified fields when articles and locations are added.


sturmer-eha:
Should be good to merge 👍 

One note, probably for a separate story: it's possible to access that template/route through http://localhost:3000/user-event/abcd. I believe hat's okay for now, but later it might be a good idea to check if such userEvent record exists before exposing that page.

PR #7, opened 2016-07-29
upgrade project's Meteor version to 1.4.0.1

sturmer-eha:
Had to comment out `standard-minifier-css@1.1.8` from .meteor/packages due to conflict with one of 3rd-party dependencies. It's a small price to pay for the latest version of Meteor, in my opinion.
Meteor 1.4 uses MongoDB 3.0 on the server and we'll have to run `meteor reset` in order to upgrade the local db to utilize all the new features that it enables us to use.

If having no minified CSS is an issue we could hold the update until that 3rd-party package gets fixed to work under 1.4.

! upd:
Also due to the upgrade of MongoDB driver other branches will need a rebase against this one in order to start local server. We'll have to upgrade all of our projects to Meteor 1.4 eventually.


nathanathan:
I applied the update in this commit due to conflicts in this branch:
https://github.com/ecohealthalliance/eidr-connect/commit/b860d0cc471b8b5f71b2c0c5f12268d65770abff

PR #8, opened 2016-07-29
Display event URL at top of event page with suggestion to link to it

mawaruDOLL:
I had a really hard time trying to fit the share link in without the page looking messy. I rearranged things so that the edit, save, and cancel buttons are at the very top of the page and a share button is now where the edit buttons used to be. Clicking the share button opens a panel where the user can copy the page's url.


nathanathan:
👍

PR #9, opened 2016-07-29
Event map is not showing up on master branch

mawaruDOLL:
I'm pretty sure I was seeing empty maps on the event pages when I shouldn't have, but things seemed to straighten out after I made a branch and checked it out. Please comment on whether you see event locations plotted on the map when an event has locations.

Otherwise, I brought back the event map menu option and fixed filtering it by event name.


nathanathan:
👍

PR #10, opened 2016-07-29
Hide EIDR C event page map when there are no locations

mawaruDOLL:
The map is now hidden when the event has no locations.


nathanathan:
👍

PR #12, opened 2016-08-01
Admins can delete EIDR C events

mawaruDOLL:
I combined the edit and delete options into a dropdown menu under a button that says "Admin Options". I don't like that the dropdown menu covers the event name when it's open, maybe we can find somewhere else to put the button. I'm not sure what the reason for an event being deleted would be so I didn't really see a need to do a soft delete at this point in time.


nathanathan:
👍

PR #13, opened 2016-08-02
Remove locations from event creation

treyyoder:
* Remove 'View Historical Events' button from the splash screen.


nathanathan:
👍

PR #16, opened 2016-08-03
Save associated articles for each location

mawaruDOLL:
Since a source article must be chosen when adding a location, I hid the location form when no articles have been added to the event yet and took away the ability to add multiple locations at once on that form. The sources are listed in the expandable location details section and can be edited there as well. When an article is removed, it's references are also removed from all locations for that event. If a location has only the removed article as its source, the location is deleted as well.


nathanathan:
👍

PR #17, opened 2016-08-03
Rewrite EIDR Connect About Page

mawaruDOLL:
I updated the text to reference the application using the name EIDR Connect rather than EIDR and to more closely reflect the capabilities of EIDR Connect. I also removed the methods and reference sections since I don't think they apply anymore. I'm assuming the funding acknowlegements will be pretty similar to EIDR, so while I put placeholder text for them, I just commented out what was there originally. I didn't update any images.


nathanathan:
👍

PR #19, opened 2016-08-08
Case death counts

treyyoder:
* Created a Case/Death Count module for the event.
* Allow 1 or more articles per count

### DIFF
--- a/collections/counts.coffee
+++ b/collections/counts.coffee
@@ -1,0 +1,51 @@
+Counts = new Meteor.Collection "counts"
+
+@grid ?= {}
+@grid.Counts = Counts
+
+getEventCounts = (userEventId) ->
+  Counts.find({userEventId: userEventId})
+
+Counts.getEventCounts = getEventCounts
+
+if Meteor.isServer
+  Meteor.publish "eventCounts", (ueId) ->
+    getEventCounts(ueId)
+
+  Counts.allow
+    insert: (userID, doc) ->
+      return true
+    remove: (userID, doc) ->
+      return Meteor.user()
+
+Meteor.methods
+  addEventCount: (eventId, url, cases, deaths, date) ->
+    if url.length
+      insertCount = {
+        url: url,
+        userEventId: eventId
+      }
+      existingCount = Counts.find(insertCount).fetch()
+      if existingCount.length is 0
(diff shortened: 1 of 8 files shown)


nathanathan:
I'm seeing a couple bugs:
 * When a count is submitted the articles field is not cleared.
 * When only a case count and source is entered, the count is not displayed.

Also, if neither count nor death count is not included, that should trigger a validation error. And, in the count list, we should only display the cases or deaths field depending on which on the user entered.

collections/counts.coffee
```diff
@@ -0,0 +1,68 @@
...
+  Counts.find({userEventId: userEventId})
+
+Counts.getEventCounts = getEventCounts
+
+if Meteor.isServer
+  Meteor.publish "eventCounts", (ueId) ->
+    getEventCounts(ueId)
+
+  Counts.allow
+    insert: (userID, doc) ->
```
nathanathan: We should check that the user is an admin.

collections/counts.coffee
```diff
@@ -0,0 +1,68 @@
...
+          # months are 0 indexed, so subtract 1 when creating the date
+          insertCount.date = new Date(dateSplit[2], dateSplit[0] - 1, dateSplit[1])
+          for location in locations
+            insertCount.location = location
+          insertCount.cases = cases
+          insertCount.deaths = deaths
+        newId = Counts.insert(insertCount)
+        Meteor.call("updateUserEventLastModified", eventId)
+        return newId
+      else
```
nathanathan: We shouldn't update existing counts.

collections/counts.coffee
```diff
@@ -0,0 +1,58 @@
...
+          insertCount.date = new Date(dateSplit[2], dateSplit[0] - 1, dateSplit[1])
+        for location in locations
+          insertCount.location = location
+        insertCount.cases = cases
+        insertCount.deaths = deaths
+        newId = Counts.insert(insertCount)
+        #Meteor.call("updateUserEventLastModified", eventId)
+        return newId
+      else
+        throw new Meteor.Error(500, "An Count for this article and event already exists.")
```
nathanathan: Articles can have multiple case counts.

nathanathan:
👍

PR #24, opened 2016-08-09
Location Sources detail popover

treyyoder:
* Add a footnote marker beside of the location source that opens a modal popover displaying the ProMED article.


nathanathan:
Could you put the footnotes next to the location name rather than in the detail menu and use the wikipedia`[1][2][3]` format.

nathanathan:
Something is breaking the location selector on this branch.

nathanathan:
👍

PR #25, opened 2016-08-10
Create test for EIDR C account creation

mawaruDOLL:
I wrote cucumber tests that will create a testing admin account, log in to the application, submit a blank user form, and submit a valid user form. I also adapted the test runner script from the mobile survey repository that automates dropping the database so the tests can be run against a clean one and restoring the original database after testing. In order to test, you will need to start EIDR Connect on port 13000, then execute run-tests.sh.


nathanathan:
Could you add instructions for installing the test dependencies and running the tests to the readme. Also, could you add the dependencies to the package.json file so they can be installed via `npm install`.

mawaruDOLL:
I added testing instructions to the readme and updated package.json. At first glance, I didn't see any npm packages that allowed running mongo commands in a terminal so I just linked the mongoDB download page but I'll keep looking for an appropriate package while testing the event form.

mawaruDOLL:
I improved running tests by adding scripts to package.json and updated the readme in #28, where the account tests are also included, so I'm closing this pull request.

PR #31, opened 2016-08-16
add timezone field to the article form, misc improvements

sturmer-eha:
improve the way we trigger the bootstrap calendar plugin
cache the date response from the promed server


nathanathan:
👍

PR #32, opened 2016-08-16
Users can see EIDR-C map markers grouped by event

mawaruDOLL:
[eventcolors.zip](https://github.com/ecohealthalliance/eidr-connect/files/420572/eventcolors.zip)

I updated the event creation process so that a color is generated from a hash of the event name. This color is then used on the event maps to show the locations associated with the event. Map markers on the event map contain different colors when there are multiple events at a single location. I tested up to 5 events at one location. The popup also lists all of the events with a small sample of the event color. The attached zip file contains a js script for updating the event collection to contain the event color in each document.

### DIFF
--- a/client/lib/mapMarkers.coffee
+++ b/client/lib/mapMarkers.coffee
@@ -7,7 +7,7 @@ Meteor.mapHelpers = {
     radius = size / 2
 
     angle = 360 * (1 / events.length)
-    angleCalc = if events.length is 1 then 1 else if (angle > 180) then 360 - angle else angle
+    angleCalc = if events.length is 1 then 0.1 else if (angle > 180) then 360 - angle else angle
     angleRad = angleCalc * Math.PI / 180
     z = Math.sqrt(2 * radius * radius - (2 * radius * radius * Math.cos(angleRad)))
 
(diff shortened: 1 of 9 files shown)


client/lib/mapMarkers.coffee
```diff
@@ -0,0 +1,32 @@
+Meteor.mapHelpers = {
```
nathanathan: It would be preferable to use export/require modules than add things to the Meteor object. Here's an example of how that works:
https://github.com/ecohealthalliance/spa/blob/master/imports/data/feeds.coffee#L40
https://github.com/ecohealthalliance/spa/blob/master/client/templates/trend_map.coffee#L1

nathanathan:
For the colors, I think we could be better to create a color ramp using https://atmospherejs.com/rcy/chromajs then evenly distribute the event colors along the ramp so we have more control over the color scheme. From a coding perspective, the hashing is cool.

mawaruDOLL:
I added chroma.js and used it to generate a gradient from which equidistant colors are extracted based on how many events are on the map. I wasn't able to come up with a way to keep the color the same on both maps without some really complex logic that didn't seem to be worth implementing, so the event color is no longer saved in the event collection. The database update script won't be necessary. 

The default gradient colors I chose probably need to be adjusted.

nathanathan:
👍

PR #33, opened 2016-08-16
case count source articles can be selected from the list of existing articles

mawaruDOLL:
While I was working with the case count article input, I made it a single select field since we don't really want multiple articles associated with one case count right now. I kept the url field an array in the counts collection so we don't have to update any existing data.


nathanathan:
👍

PR #34, opened 2016-08-18
Add legend to map

mawaruDOLL:
I limited the map to 5 events at a time and ordered them by creation date, so the most recent events will be at the top of the legend. The legend has buttons to view older and newer events and can be collapsed down in case it gets in the way of any map markers.


nathanathan:
👍 Looks good

PR #35, opened 2016-08-18
Case Counts / Incident Reports can have multiple locations

mawaruDOLL:
I pulled the location select2 code out into its own template so it can be shared between the location and case count form and made the input on the case count form a multiple select input. The counts collection will need to be updated so that the locations property is an array, which the attached script will take care of.

[updatecountlocations.zip](https://github.com/ecohealthalliance/eidr-connect/files/425558/updatecountlocations.zip)


nathanathan:
👍 This looks good. I'm going to wait until we have backups in place to deploy this because I am cautious about making changes to the database without them. For future PRs that have accompanying scripts, could you put the scripts in the .scripts folder.

PR #36, opened 2016-08-19
Change case counts to incident reports

mawaruDOLL:
I added a field for selecting the incident type, which toggles the following form element between a number input and a text input. In order to avoid database changes, I kept the case count, death count, and other specification values as separate document properties in the collection rather than having type and value fields.

### DIFF
--- a/client/views/userEvent.jade
+++ b/client/views/userEvent.jade
@@ -259,7 +259,7 @@ template(name="articleSelect2")
 template(name="counts")
   .panel.panel-default
     .panel-heading
-      h3 Case/Death Counts
+      h3 Incident Reports
     .panel-body.comments-container
       if currentUser
         .comment-form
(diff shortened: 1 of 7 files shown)


collections/counts.coffee
```diff
@@ -20,7 +20,7 @@ if Meteor.isServer
       return Roles.userIsInRole(Meteor.userId(), ['admin'])
 
 Meteor.methods
-  addEventCount: (eventId, url, locations, cases, deaths, date) ->
+  addEventCount: (eventId, url, locations, type, count, date) ->
```
nathanathan: This should be renamed to addIncidentReport and the count argument removed.

client/views/userEvent.jade
```diff
@@ -277,48 +277,64 @@ template(name="counts")
...
+              | No location specified.
+      .row
+        if cases
+          +incident label="Cases" value=cases
+        if deaths
+          +incident label="Deaths" value=deaths
+        if specify
+          +incident label="Other" value=specify
+
+template(name="incident")
```
nathanathan: How about incidentType?

client/views/userEvent.jade
```diff
@@ -277,48 +277,64 @@ template(name="counts")
...
+                button.btn.btn-primary.btn-block(type="submit") Add Report
     if counts.length
       ul.list-group
         each counts
           +count
     else
-      p No case/death counts are associated with this event.
+      p No incident reports are associated with this event.
 
 template(name="count")
```
nathanathan: Could you rename this template to incidentReport and put it in a separate file?

nathanathan: The same applies to the counts template.

nathanathan:
👍

PR #37, opened 2016-08-19
Use geoname-lookup.eha.io to find geonames

nathanathan:


mawaruDOLL:
This looks good. The only issue I found is that the links to geonames.org in the location list on the event page don't work properly since the addition of crossorigin.me to the url, but it isn't a change introduced with this pull request.

PR #38, opened 2016-08-19
Events can have a disease associated with them

mawaruDOLL:
I didn't add a disease field to the new event form because I'm thinking that if we do any autocomplete/diagnosis in the future, it will probably be based on the articles which are added to the event and these can't be added on the new event form. I had trouble deciding where to place the disease on the event page. I think it looks rather out of place as is.


nathanathan:
👍

PR #39, opened 2016-08-23
Events on map can be filtered by dates of case counts

mawaruDOLL:
The filter control on the event map now contains options for filtering by incident report date.


nathanathan:
When filtering by date, I see events show up in the legend, but they do not appear on the map.

imports/ui/dateHelpers.coffee
```diff
@@ -0,0 +1,12 @@
+dateStringToDate = (dateString, delim) ->
```
nathanathan: Could you replace the dateHelpers module with moment.js? Calls like `moment(dateString, "MM/DD/YYYY")` or `moment(date).format("MM/DD/YYYY")` are preferable since it possible to tell what kind of format is being used without referring to a function definition in another file.

mawaruDOLL:
The events that appeared in the legend but not on the map just didn't have any locations, so I excluded them from the legend. I also included moment.js.

PR #40, opened 2016-08-23
Apply EIDR connect header styles from mockups

mawaruDOLL:
I added the new logo and updated the primary and secondary style colors. I think the highlight color for the menu options makes the white text a little hard to read but I went along with the existing rules that made this color the same as the border color.


nathanathan:
I agree about the highlight color. Maybe we could make the text color change to the secondary color on highlight to compensate.
A couple other issues I noticed:
 * The logo is getting squished due to its width being hard-coded.
 * The splash page needs more space above the logo.
 * In the mockups, the text color for the header buttons is a bit lighter.

mawaruDOLL:
I fixed the issues listed in the previous comment.

PR #43, opened 2016-08-25
Move Incident Reports and Location panels into tabs

mawaruDOLL:
I'm not sure this is ready to merge as is but I'd like some feedback.

Known issues:
* The map is still using the locations collection and is broken for some events.
* Opening a location detail row and reordering the rows doesn't change the position of the detail row.
* Adding and deleting incident reports doesn't close or re-position open location detail rows.


client/controllers/locationSelect2.coffee
```diff
@@ -34,4 +35,5 @@ Template.locationSelect2.onRendered ->
               item: hit._source
             }
         }
-  @$(".select2-container").css("width", "100%")
+  #override the inline width of the select2 container and its text input field
+  $input.next(".select2-container").css("width", "100%").find("input").css("width", "100%")
```
nathanathan: I think this could be done with css

nathanathan:
This looks great so far. The incident reports form should be moved to a modal that opens when an "Add incident report" button is clicked. Also, could you add search boxes using this reactive table feature: https://github.com/aslagle/reactive-table#multiple-filters-outside-a-table

client/controllers/locationSelect2.coffee
mawaruDOLL: I did it this way because the widths are set directly on these elements as inline styles. I'm not sure if a css rule would override that.

mawaruDOLL:
I added a filter outside of the incident report table but all of the filterable columns are virtual columns. The virtual columns section of the documentation points out that filtering on these types of columns isn't possible at the moment.

mawaruDOLL:
I fixed the detail sections being stuck in the table after reordering by using toHTML() rather than render(), so keeping track of the current view for destruction purposes isn't necessary. Now, when the incident tab is opened or the location table is reordered, open detail rows are removed manually. I kept the location table filter since it will work with the name column but I was unable to add one for incident reports. Also, the incident report form now opens in a modal.

nathanathan:
👍 Looks good

PR #44, opened 2016-08-25
prevent clicks from going through the legend container

sturmer-eha:
This PR prevents the clicks on the legend widget from going through to the map.
Also adds a warning toast when the user attempts to add an article which is already in the database.


nathanathan:
👍

PR #45, opened 2016-08-26
Show locations from incident reports on map

mawaruDOLL:
I updated the event map and event page map to plot locations from incident reports rather than the location collection.

### DIFF
--- a/client/controllers/eventMap.coffee
+++ b/client/controllers/eventMap.coffee
@@ -56,8 +56,8 @@ Template.eventMap.onRendered ->
 
     if _.isObject query
       allEvents = instance.data.events.find(query, {sort: {creationDate: -1}}).fetch()
-      startingPosition = currentPage * instance.eventsPerPage
       totalEventCount = allEvents.length
+      startingPosition = currentPage * instance.eventsPerPage
     else
       map.removeLayer(markers)
       return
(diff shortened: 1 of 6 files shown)


client/controllers/userEvent.coffee
```diff
@@ -14,6 +14,26 @@ Template.userEvent.onRendered ->
 Template.userEvent.helpers
   isEditing: ->
     return Template.instance().editState.get()
+  incidentLocations: ->
+    locations = {}
+    counts = Template.instance().data.counts
+    # Loop 1: Incident Reports ("counts")
+    for count in counts
+      # Loop 2: Locations within each "count" record
+      for loc in count.locations
```
sturmer-eha: There was a bug in this helper, the fixes are available back in the master branch, file locationList.coffee
Namely, there has to be one more line to check if that `count` has any `.locations` has and `date.parse` should be `Date.parse`

mawaruDOLL: Ok, I'll introduce the fixes. But just as a side note, this helper won't be staying in the userEvent controller for very long. The map is going to be moved to the locationList template in #43.

client/controllers/eventMap.coffee
```diff
@@ -59,33 +59,44 @@ Template.eventMap.rendered = ->
...
+      filteredEvents = []
       colorScale = chroma.scale(MapHelpers.getDefaultGradientColors()).colors(instance.eventsPerPage)
-
-      while templateEvents.length < instance.eventsPerPage and eventIndex < totalEventCount
-        event = allEvents[eventIndex]
-        eventLocations = instance.data.locations.find({userEventId: event._id}).fetch()
-
-        if eventLocations.length
+      
+      #Remove events that have no locations to plot on the map
```
sturmer-eha: It's way easier to read verbose comments when they have a space after the comment symbol.
I found it to be useful to comment code out without a space and put a space and make the first letter capital when comment contains natural language.

nathanathan:
:+1:

PR #47, opened 2016-08-26
Add travel-related checkbox to incident report

mawaruDOLL:
I added the travel related field to the incident report form and details section.


mawaruDOLL:
I added this field to #48 to avoid conflicts.

PR #48, opened 2016-08-26
Add species field to incident report form

mawaruDOLL:
The species field is a non-required field and it's only displayed in the details section if it was entered.

### DIFF
--- a/client/controllers/incidentReports.coffee
+++ b/client/controllers/incidentReports.coffee
@@ -37,6 +37,12 @@ Template.incidentReports.helpers
         cellClass: "remove-row"
       })
 
+    fields.push({
+      key: "expand"
+      label: ""
+      cellClass: "open-row"
+    })
+
     return {
       id: 'event-incidents-table'
       fields: fields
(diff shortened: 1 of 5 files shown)


client/controllers/incidentReports.coffee
```diff
@@ -67,7 +67,7 @@ Template.incidentReports.events
 
     incidentCount = if e.target.count then e.target.count.value.trim() else e.target.other.value.trim()
 
-    Meteor.call("addIncidentReport", templateInstance.data.userEvent._id, article, allLocations, e.target.incidentType.value, incidentCount, e.target.date.value, (error, result) ->
+    Meteor.call("addIncidentReport", templateInstance.data.userEvent._id, article, allLocations, e.target.incidentType.value, incidentCount, e.target.date.value, e.target.species.value, (error, result) ->
```
nathanathan: I think it would be better to pass in this data as properties of a single object instead given the large number of arguments.

nathanathan:
Could you resolve the conflicts in this and #47 ? It would help to combine them into a single PR since they will probably conflict with each other.

mawaruDOLL:
I resolved the conflicts and combined the travel related field from #47 into this pull request. Since it was difficult to make sure the new fields were being saved correctly, I added an expandable incident report detail section to the table. I'm sure the layout will need to be adjusted but it displays the necessary information for now.

nathanathan:
:+2: Great work! I like the expandable incident reports.

PR #49, opened 2016-08-26
show relevant locations by default in the location omnibox

sturmer-eha:
We may want to sort the items alphabetically later, or perhaps by relevance.
The users have also suggested removing the tag from the input completely, bringing back whatever was typed in prior selecting a location.
Both qualify to be separate stories, in my opinion.


nathanathan:
When I tried to merge this with the master branch, the article selector broke, and the location selector options were showing up behind the modal. The location input also lost its styling.

sturmer-eha:
It should be fine now, please take a look to see whenever you get a minute.

nathanathan:
This is working pretty well, but I think we should find another way to debounce the typeahead. The added load makes it take about twice as long to show a correct set of suggestions and there is a lot of jitteriness while it's rendering the intermediate results.

sturmer-eha:
I could debounce the whole callback altogether, not just the ajax call. But it will create a delay on rendering data that's available within the browser already (mongodb). Let me try that and let's see if it's better...

nathanathan:
👍

PR #50, opened 2016-08-29
Rename eligable count variables to incident report

mawaruDOLL:
I updated the remaining counts variables so the naming reflects that they are related to incident reports.


nathanathan:
👍

PR #52, opened 2016-08-29
Entering a location search term should close expanded rows

mawaruDOLL:
I added a keyup event handler to the location filter that removes detail rows when a search term is entered.


nathanathan:
👍

PR #53, opened 2016-08-30
Incident report / location tab navigation should be based on urls instead of events

mawaruDOLL:
The initial event page load still uses the same route and url. Clicking the tabs will change between 2 new routes I added. It's annoying how the page jumps to the top after clicking a tab but I'm not sure of how to keep that from happening.

### DIFF
--- a/client/controllers/userEvent.coffee
+++ b/client/controllers/userEvent.coffee
@@ -15,13 +15,13 @@ Template.userEvent.helpers
   isEditing: ->
     return Template.instance().editState.get()
   locationView: ->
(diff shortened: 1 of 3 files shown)


nathanathan:
The route patterns should be subpaths of the event page, for instance "/user-event/:_id/locations." We could deal with the scroll position by creating an autorun function on the template that sets the scroll position to the tabs section when an incident or location path is passed in, but that can be handled in a future story.

mawaruDOLL:
I fixed the tab urls so that they are subpaths of the event page.

nathanathan:
👍

PR #54, opened 2016-08-30
Update db to split case/death counts into separate incidents

mawaruDOLL:
I added a startup method that splits incident reports with both case and death counts into 2 incidents.

### DIFF
--- a/collections/incidents.coffee
+++ b/collections/incidents.coffee
@@ -62,7 +62,7 @@ if Meteor.isServer
   Meteor.startup ->
     incidents = Incidents.find({$and: [{cases: {$exists: true}}, {deaths: {$exists: true}}]}).fetch()
     for incident in incidents
-      if incident.cases.length and incident.deaths.length
+      if incident.cases?.length and incident.deaths?.length
         newIncident = {
           url: incident.url
           userEventId: incident.userEventId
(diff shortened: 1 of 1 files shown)


collections/incidents.coffee
```diff
@@ -55,3 +55,28 @@ Meteor.methods
       Incidents.remove(id)
       #Meteor.call("removeOrphanedLocations", removed.userEventId, id)
       Meteor.call("updateUserEventLastModified", removed.userEventId)
+
+# Split incidents with both case and death counts into separate incidents
+if Meteor.isServer
+  Meteor.startup ->
+    incidents = Incidents.find({$and: [{cases: {$exists: true}}, {deaths: {$exists: true}}]}).fetch()
+    for incident in incidents
+      if incident.cases.length and incident.deaths.length
```
sturmer-eha: I would do incident.cases?.length and incident.deaths?.length here.

collections/incidents.coffee
```diff
@@ -55,3 +55,28 @@ Meteor.methods
...
+          addedDate: incident.addedDate
+          deaths: incident.deaths
+        }
+
+        if incident.locations?.length
+          newIncident.locations = incident.locations
+        if incident.date
+          newIncident.date = incident.date
+
+        Incidents.update({_id: incident._id}, {$unset: {deaths: ""}})
```
sturmer-eha: You could do just `.update(incident._id, {$unset: ...` here instead, but it's just to make it more compact, no difference otherwise.

mawaruDOLL:
I implemented the requested code improvements.

PR #55, opened 2016-08-30
Make incident report table the default tab and change columns

mawaruDOLL:
The order of the tabs has been switched and the extra columns added to the incident reports table.

### DIFF
--- a/client/controllers/incidentReports.coffee
+++ b/client/controllers/incidentReports.coffee
@@ -23,10 +23,12 @@ Template.incidentReports.helpers
           return ""
       },
       {
-        key: "addedDate"
-        label: "Reported"
+        key: "date"
+        label: "Date"
         fn: (value, object, key) ->
-          return moment(value).fromNow()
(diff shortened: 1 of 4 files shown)


client/controllers/incidentReports.coffee
```diff
@@ -27,6 +27,18 @@ Template.incidentReports.helpers
...
         fn: (value, object, key) ->
           return moment(value).fromNow()
+      },
+      {
+        key: "travelRelated"
+        label: "Travel Related"
+        fn: (value, object, key) ->
+          if value
+            return "Yes"
+          return "No"
```
nathanathan: When the travel related box isn't checked, we should just leave fields like this blank instead of interpreting it as meaning not travel related. The check box could be empty because the travel status is unknown or because it is an old record from before the field was added.

nathanathan:
The reported column should be replaced with the incident date.

nathanathan:
👍

PR #57, opened 2016-08-30
When there are no incident reports for an event, show "create incident report" button instead of tabs

mawaruDOLL:


nathanathan:
👍

PR #58, opened 2016-08-31
Add article count column to tracked events table

mawaruDOLL:
I tried to get this to work without adding an extra field to the event collection but it just wouldn't work out. So I went ahead and added a field that holds the article count which gets updated when articles are added and deleted.

### DIFF
--- a/collections/userEvents.coffee
+++ b/collections/userEvents.coffee
@@ -63,12 +63,13 @@ Meteor.methods
         lastModifiedByUserName: user.profile.name
       }})
   updateUserEventArticleCount: (id, countModifier) ->
-    event = UserEvents.findOne(id)
-    UserEvents.update(id, {$set: {articleCount: event.articleCount + countModifier}})
+    if Meteor.user()
+      event = UserEvents.findOne(id)
+      UserEvents.update(id, {$set: {articleCount: event.articleCount + countModifier}})
 
 # Add article count field to events
 if Meteor.isServer
   Meteor.startup ->
-    UserEvents.find({articleCount: {$exists: false}}).forEach (event) ->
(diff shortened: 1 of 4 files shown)


collections/userEvents.coffee
```diff
@@ -61,3 +62,13 @@ Meteor.methods
         lastModifiedByUserId: user._id,
         lastModifiedByUserName: user.profile.name
       }})
+  updateUserEventArticleCount: (id, countModifier) ->
+    event = UserEvents.findOne(id)
```
nathanathan: We should check that the user is an admin.

collections/userEvents.coffee
```diff
@@ -61,3 +62,13 @@ Meteor.methods
...
+    UserEvents.update(id, {$set: {articleCount: event.articleCount + countModifier}})
+
+# Add article count field to events
+if Meteor.isServer
+  Meteor.startup ->
+    UserEvents.find({articleCount: {$exists: false}}).forEach (event) ->
```
nathanathan: I would remove filter query and just update the counts on all the events incase one gets out of sync somehow.

collections/userEvents.coffee
mawaruDOLL: It seems that any user can add articles to an event right now, not just admins. Should I check for a user instead?

nathanathan: I think it would be better to require the admin role incase we want to add a type of account that doesn't have editing privileges.

mawaruDOLL:
I updated the necessary templates and server methods so only admins are allowed to add and remove articles.

nathanathan:
👍

PR #59, opened 2016-08-31
Update article deletion

mawaruDOLL:
I added a toastr error message that displays when attempting to delete an article that has associated incident reports and updated the confirmation message when an article can be deleted.

### DIFF
--- a/client/controllers/article.coffee
+++ b/client/controllers/article.coffee
@@ -1,9 +1,11 @@
 Template.article.events
   "click .delete-article": (e) ->
-    incidentCount = Incidents.find({url: @url}).count()
+    incidentCount = Incidents.find({userEventId: @userEventId, url: @url}).count()
     if incidentCount
       plural = if incidentCount is 1 then "" else "s"
-      message = "There " + (if incidentCount is 1 then "is an incident report" else "are #{incidentCount} incident reports") + " associated with this article. Please delete the incident report#{plural} before deleting the article."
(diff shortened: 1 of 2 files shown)


client/controllers/article.coffee
```diff
@@ -1,4 +1,9 @@
 Template.article.events
   "click .delete-article": (e) ->
-    if window.confirm("Are you sure you want to delete this article?\nAll locations mentioned only in this article will be deleted as well.")
+    incidentCount = Incidents.find({url: @url}).count()
+    if incidentCount
+      plural = if incidentCount is 1 then "" else "s"
+      message = "There " + (if incidentCount is 1 then "is an incident report" else "are #{incidentCount} incident reports") + " associated with this article. Please delete the incident report#{plural} before deleting the article."
```
nathanathan: Could you break this into multiple lines.

client/controllers/article.coffee
```diff
@@ -1,4 +1,9 @@
 Template.article.events
   "click .delete-article": (e) ->
-    if window.confirm("Are you sure you want to delete this article?\nAll locations mentioned only in this article will be deleted as well.")
+    incidentCount = Incidents.find({url: @url}).count()
```
nathanathan: Could you change the query to only count the incidents for the current event. The subscription is probably only returning incidents for the current event, but that could change.

nathanathan:
👍

PR #60, opened 2016-09-01
Event source section

treyyoder:
* Added the Event Source Section.
* Added modal form for adding Sources.
* Updated UI to match mockups.


client/views/userEvent.jade
```diff
@@ -35,7 +35,7 @@ template(name="userEvent")
                 .col-sm-2.space-btm-1.space-top-1
                   if isInRole "admin"
                     .dropdown
-                      button.btn.secondary-btn.btn-block.dropdown-toggle(type="button" data-toggle="dropdown")
+                      button.btn.secondary-btn.btn-block.dropdown-toggle(type="button" data-toggle="dropdown" style="white-space: normal;")
```
sturmer-eha: I would try to avoid the `style=""` attribute and go via CSS with `!important`. Neither is perfect though.

collections/articles.coffee
```diff
@@ -44,7 +44,8 @@ if Meteor.isServer
 
 Meteor.methods
   addEventSource: (source) -> #eventId, url, publishDate, publishDateTZ
-    if source.url.length
+    user = Meteor.user()
+    if Roles.userIsInRole(user._id, ['admin']) and source.url.length
```
sturmer-eha: I see now that this method is both server-side and client-side, so if somebody calls that method via JS console while not being authorized, it'll trigger an exception on the server.
Might be out of scope of this PR, but I'd add a check between L47 and L48 to make sure the user is not null.

sturmer-eha:
Really well done, I'd say we should go ahead and merge 👍

PR #61, opened 2016-09-01
update to Meteor 1.4.1.1, add bcrypt NPM as a dependency

sturmer-eha:
This is a minor version bump.
Meteor now optionally uses bcrypt npm which is a binary implementation, said to be up to 3x faster than the built-in JS one.
```
(STDERR) Note: you are using a pure-JavaScript implementation of bcrypt.
(STDERR) While this implementation will work correctly, it is known to be
(STDERR) approximately three times slower than the native implementation.
(STDERR) In order to use the native implementation instead, run
(STDERR)          
(STDERR)   meteor npm install --save bcrypt
(STDERR)          
(STDERR) in the root directory of your application.
```
I've added it to the package.json file but if it breaks the build we could revert back to using the JS implementation instead.


nathanathan:
👍

PR #62, opened 2016-09-02
Event summary section

treyyoder:
Depends on #60 
* Updated the User Event Summary section to match https://drive.google.com/file/d/0B2aeANKs4QcPbjNKSk9nT1dBTXM/view


nathanathan:
There is a css conflict causing the event name column in the tracked events table to be incorrectly formatted. Also, the text "case counts" should be changed to "incident reports" and the location count should not be included.

nathanathan:
:+1:

PR #66, opened 2016-09-06
Add delete button to event sources.

treyyoder:


nathanathan:
This should replace the code here and include its check for connected incidents: https://github.com/ecohealthalliance/eidr-connect/blob/19eb5313b9a8e0216bc700aac3c82c39cbc70ecb/client/controllers/article.coffee#L3

nathanathan:
👍

PR #67, opened 2016-09-07
Stylus variables

jgoley:
Preprends stylus variables with `$`.
Cleans up stylus — removes commas from lists of selectors.


nathanathan:
👍

PR #68, opened 2016-09-07
Change incident dates to datetime ranges

mawaruDOLL:
The mockup shows an inline calendar for selecting the date ranges. The best date range library I found doesn't have an inline option but I was able to make it work. I made a fork of the original library to add the option but the feature isn't complete yet.

### DIFF
--- a/client/views/incidentForm.jade
+++ b/client/views/incidentForm.jade
@@ -1,46 +1,88 @@
 template(name="incidentForm")
   form#add-incident.form-horizontal(novalidate)
     .form-group
-      label.col-lg-3.control-label Article URL
-      .col-lg-9
-        +articleSelect2 selectId="articleSource" articles=articles selected=incidentData.articleSource
+      .col-md-12
+      label.control-label Article URL
+      +articleSelect2 selectId="articleSource" articles=articles selected=incidentData.articleSource
     .form-group
-      label.col-lg-3.control-label Date
-      .col-lg-9
-        //input.form-control.datePicker(type="text" name="date" data-DateTimePicker value=incidentData.date)
-        .form-group
-          label.col-lg-3.control-label Date
-          .col-lg-3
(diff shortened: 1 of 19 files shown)


imports/ui/inlineDateRangePicker.coffee
```diff
@@ -0,0 +1,26 @@
+createInlineDateRangePicker = (template, parentSelector, singleDatePicker) ->
```
nathanathan: To make this more generic, I suggest changing the function signature to:
```
(parentElement: jQueryWrappedElement, options: {
  startDate: optionalDate,
  endDate: optionalDate,
  singleDate: optionalBoolean
}) ->
```

imports/schemas/incidentReport.coffee
```diff
@@ -28,9 +28,17 @@ IncidentReportSchema = new SimpleSchema(
     blackbox: true
   userEventId:
     type: String
-  date:
+  dateRangeType:
```
nathanathan: I think a nested object would be better for this. For example:
```
dateRange:
  type: ...
  start: ...
  end: ...
```

client/views/incidentForm.jade
```diff
@@ -1,50 +1,68 @@
...
+            a(href="#singleDate" aria-controls="single Date" role="tab" data-toggle="tab") Single Date
+          li(role="presentation")
+            a(href="#preciseRange" aria-controls="date range" role="tab" data-toggle="tab") Date Range
+
+        .tab-content
+          #singleDate.tab-pane.active(role="tabpanel")
+            #singleDatePicker.inlineRangePicker
+            .form-group
+              .col-md-12.text-center
+                .checkbox
```
nathanathan: Could you move the cumulative checkbox to be next to the count value input.

nathanathan:
👍

PR #70, opened 2016-09-07
In tracked events table, show event name column first, and format dates to be in user's timezone instead of showing timezone offsets

mawaruDOLL:


nathanathan:
👍

PR #71, opened 2016-09-07
Update string counts in database

mawaruDOLL:


collections/incidents.coffee
```diff
@@ -67,4 +67,13 @@ if Meteor.isServer
...
\ No newline at end of file
+        Incidents.insert(newIncident)
+
+    # Convert string case and death counts to integers
+    incidents = Incidents.find({$or: [{cases: {$type: "string"}}, {deaths: {$type: "string"}}]}).fetch()
+    for incident in incidents
+      if incident.cases
+        updateField = {cases: parseInt(incident.cases)}
+      else
+        updateField = {deaths: parseInt(incident.deaths)}
```
nathanathan: The incident could be a "specify other" in this case.

nathanathan: Also, I think some error handling is needed in case parseInt doesn't return a number.

mawaruDOLL:
I accounted for parseInt() returning NaN by setting the specify field with the value that failed to parse.

collections/incidents.coffee
```diff
@@ -67,4 +67,22 @@ if Meteor.isServer
...
+          mongoProjection = {$set: {specify: incident.cases}, $unset: {cases: ""}}
+        else
+          mongoProjection = {$set: {cases: parsed}}
+      else if incident.deaths
+        parsed = parseInt(incident.deaths)
+        if parsed.toString() is "NaN"
+          mongoProjection = {$set: {specify: incident.deaths}, $unset: {deaths: ""}}
+        else
+          mongoProjection = {$set: {deaths: parsed}}
+      if mongoProjection
```
nathanathan: I think the mongoProjection variable form the last iteration might still be defined.

nathanathan:
👍

PR #72, opened 2016-09-07
Move about page content

jgoley:
I ended up just removing most of the about content for now. We need to draft some quality informational copy that communicates the significance of the application.

- Moves away from PNGs in favor of SVGs (replaces EIDR-C and EHA logos)
- Updates the favicon
- Customizes button styling


nathanathan:
👍

PR #73, opened 2016-09-07
Add species and confirm status to incident report expanded table rows, and when travel related is not true, leave it blank instead of showing an X

mawaruDOLL:
The status and species fields were already being displayed in the incident details but only if they had values, so I took away the if that kept them from being shown when they were blank. And if the incident isn't travel related, the travel related field is blank rather than containing X.


nathanathan:
👍

PR #76, opened 2016-09-08
Store locations using schema form geoname lookup service

mawaruDOLL:
I updated existing incident locations so the property names match the schema definition, so I also had to adjust a few places in the code that referenced these properties.


client/controllers/incidentReport.coffee
```diff
@@ -1,10 +1,14 @@
+formatLocation = require '/imports/formatLocation.coffee'
+
 Template.incidentReport.helpers
   travelIcon: ->
     if @travelRelated
       return "fa-check"
     return "fa-times"
   formatDate: (date) ->
     return moment(date).format("MMM D, YYYY")
+  formatLocation: (location) ->
```
nathanathan: There is a global template helper for this.

client/controllers/incidentModal.coffee
```diff
@@ -22,8 +22,6 @@ Template.incidentModal.events
            $(form.count).val("")
            $(form.incidentType).val(null).trigger("change")
            $(form.travelRelated).attr('checked', false)
-        if closeModal
```
nathanathan: Why was this deleted?

imports/schemas/incidentReport.coffee
```diff
@@ -40,5 +37,32 @@ IncidentReportSchema = new SimpleSchema(
   status:
     type: String
     optional: true
+  "locations":
```
nathanathan: This key is defined twice.

client/controllers/incidentModal.coffee
mawaruDOLL: I was getting an error because closeModal wasn't defined. Should the modal be closed after "save" is clicked?

nathanathan: No, it looks like a left over from before we changed the modal buttons. You're right to delete it.

collections/incidents.coffee
```diff
@@ -53,11 +53,32 @@ Meteor.methods
...
+              countryName: loc.countryName
+              id: loc.geonameId
+              latitude: loc.latitude
+              longitude: loc.longitude
+              name: loc.name
+            )
+          else
+            updatedLocations.push(loc)
+        if updatesMade
+          Incidents.update(incident._id, {$set: locations: {updatedLocations}})
```
nathanathan: I think updatedLocations should be in an array.

nathanathan:
The geonameId property is still used in some places:
```
/eidr-connect/collections/geolocations.coffee:
	18:         existingLocations.push(loc.geonameId)
	24:         if existingLocations.indexOf(location.geonameId.toString()) is -1
	28:             geonameId: location.geonameId,
	98:           geonameId: geolocation.geonameId

/eidr-connect/collections/incidents.coffee:
	65:           if loc.geonameId
	70:               id: loc.geonameId
```

mawaruDOLL:
Does the geolocations collection need to be updated to use the schema as well even though it has been deprecated?

The references to geonameId in collections/incidents.coffee are in the part of the startup function that is updating locations in the counts collection to match the location schema so I don't think these lines need to be changed.

nathanathan:
That code doesn't need to be updated, but since we're no longer using it could you delete it?

imports/schemas/incidentReport.coffee
```diff
@@ -43,5 +40,30 @@ IncidentReportSchema = new SimpleSchema(
...
+    optional: true
+  "locations.$.countryName":
+    type: String
+  "locations.$.featureClass":
+    type: String
+  "locations.$.featureCode":
+    type: String
+  "locations.$.id":
+    type: String
+  "locations.$.latitude":
```
nathanathan: latitude and longitude should be numbers

client/controllers/suggestedIncidents.coffee
```diff
@@ -58,13 +58,17 @@ Template.suggestedIncidentsModal.onCreated ->
...
-          geonameId: loc.id
+          id: loc.id
           name: loc.name
-          displayName: loc.name
-          subdivision: loc.admin1Name
-          latitude: parseFloat(loc.latitude)
-          longitude: parseFloat(loc.longitude)
+          admin1Name: loc.admin1Name
+          admin2Name: loc.admin2Name
+          latitude: loc.latitude
```
nathanathan: The parseFloat function should be kept

nathanathan:
There is a new geonameId reference that was added here: https://github.com/ecohealthalliance/eidr-connect/blob/master/client/controllers/articles.coffee#L60

nathanathan:
👍

PR #77, opened 2016-09-08
Set up Continuous Integration / Docker

sjmatta:
* Requires a `settings-production.json` in the root directory
* Currently exposes the Meteor application directly, without a reverse proxy
* Added native bcrypt install - this may be more appropriately placed in `package.json`
* Looking for additional feedback


sjmatta:
* Add `circle.yml` to build Docker container and archive as an artifact
* (Also archives docker-compose config for convenience)
* Testing is not part of continuous integration, yet

sjmatta:
Finishes #120003861

sjmatta:
* Tests are now run
* All tests are currently passing, but, I can't figure out why Chimp (or, at least, run_tests.sh) isn't returning a non-zero code when tests fail

sjmatta:
Unfortunately, the build isn't failing when tests fail. I'll work on this some more if I have additional time, but I think given that as known technical debt, this is ready to merge.

nathanathan:
Could you create a separate Dockerfile for CI and testing. The one you've been editing is currently used for deployments, so I don't want to merge all of these changes to it since some will probably break our deployment process.

@frosario Could you comment on this as well.

Dockerfile
```diff
@@ -1,38 +1,28 @@
-FROM ubuntu:14.04.4
+FROM ubuntu:14.04
```
frosario: Why base off an older obsolete release of ubuntu?

Dockerfile
```diff
@@ -1,38 +1,28 @@
...
 
-# Replace shell with bash so we can source files
-RUN rm /bin/sh && ln -s /bin/bash /bin/sh
```
frosario: This line guarantees that when you run an exec command, that we are using the bash shell and not the minimal dash shell. Bash is necessary for things like properly sourcing in environment vars.

Dockerfile
sjmatta: According to https://hub.docker.com/_/ubuntu/, :14.04 will always point to the latest release of ubuntu 14.04 (currently it points to 14.04.5).

Dockerfile
```diff
@@ -1,38 +1,28 @@
...
-
-# Install apt package dependencies
-RUN apt-get clean all && apt-get update && \
-    apt-get -y install wget curl python make g++ git supervisor && \
```
frosario: We need supervisor as the process that keeps the container alive always. Typical docker containers are supposed to be single process, but we had to take a very hybrid approach for our stakeholder's needs.

Dockerfile
sjmatta: Oh, I may have misunderstood. You mean to use 16?

Dockerfile
sjmatta: @nathanathan has asked I make this Dockerfile separate from the existing one, for use in dev. This one won't be used to create containers delivered to your stakeholders.

Dockerfile
frosario: Correct. 16 has been out for awhile now and seems pretty stable. Any reason to stay on 14?

Dockerfile
frosario: Ok sounds good.

nathanathan:
I'm closing this and recreating the changes in new files on another branch.

PR #81, opened 2016-09-09
Use expanded date selector for source dates

mawaruDOLL:
The date picker is now inline and the time selector has been separated into its own input.


nathanathan:
👍 This PR looks good but need to be rebased.

mawaruDOLL:
Rebased.

PR #83, opened 2016-09-13
Show article snippets in incident suggestion modal

mawaruDOLL:
I went with displaying a specified number of characters before and after the incident link text. While testing, I noticed that dates for the suggested incidents aren't being set properly on the form. I thought it would be a quick fix and took a look at it but I don't understand the date territories well enough.


mawaruDOLL:
I had the wrong idea about this and had the article snippets being displayed in place of the full article. I fixed things so the snippet displays at the top of the suggested incident form and the complete article is still shown.

nathanathan:
👍

PR #84, opened 2016-09-13
Event map ui updates

jgoley:
This PR establishes a new UI for the event map and allows users to select individual or multiple events.

Date filtration is not working yet. If necessary, we can hide the toggle icon until its implemented.

I'm not sure if there are issues with my local data or if there is a bug somewhere but the the pagination fails on the what should be the last previous page of events - the previous button does not change state to disabled as it should. There is also a decent delay when a user switches between pages of events. I can look into this further before merging.

PT Chore: https://www.pivotaltracker.com/story/show/130284429


client/main.styl
```diff
@@ -0,0 +1 @@
+@import '{}/imports/stylesheets/_main'
```
nathanathan: This is causing an error for me:
```
=> Errors prevented startup:                  
   
   While processing files with stylus (for target web.browser):
   client/main.styl: Stylus compiler error: {}/imports/stylesheets/_main.styl:1:10
   1| @import '{}/imports/stylesheets/_main'
   ---------------^
   2| 
   
   failed to locate @require file ../node_modules/rupture/rupture/index.styl
```

nathanathan: Never-mind, I see an npm install is required.

nathanathan:
This looks awesome! I'll go ahead and merge now to try and reduce conflicts from all the refactoring. One tweak I think we should make is to reverse the direction of the pagination arrows.

jgoley:
@nathanathan I think we should probably decrease the number of events in the list to between 5 and 10 (It was previously 5 which I thought was on the low side). 10 events makes distinguishing the colors of the markers more difficult and on smaller screens the list tends to overflow the viewport.

PR #85, opened 2016-09-14
Reduce number of event list items and improve styling

jgoley:
Reduces items in event map list to 8 from 10.
Also tweaks layout on wide screens.


nathanathan:
👍

PR #86, opened 2016-09-14
Fix styling and text-sive of buttons

jgoley:
Fixes up the customized set of bootstrap button classes.
Adds `.btn-delete` as alias for the `.btn-danger` and `.btn-confirm` for the `.btn-success`
Adds a few classes for more control over the styling of buttons:
`.btn-spacious` increases padding and font-size
`.btn-action` uppercases text (alias: `.shouting`)

Also removes `.main-btn` and `.secondary-btn` as the styles for those classes were removed at some point. Now we should use bootstraps btn syntax.

https://www.pivotaltracker.com/story/show/130077331

### DIFF
--- a/client/controllers/userEvent.coffee
+++ b/client/controllers/userEvent.coffee
@@ -6,7 +6,7 @@ Template.userEvent.onCreated ->
 
 Template.userEvent.onRendered ->
   $(document).ready ->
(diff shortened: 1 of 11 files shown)


nathanathan:
👍

PR #87, opened 2016-09-14
Footer update

treyyoder:
* Updated the footer to match mockups.


client/views/footer.jade
```diff
@@ -1,5 +1,15 @@
...
+      .pull-right
+        a(href="mailto:eidr@ecohealthalliance.org") eidr@ecohealthalliance.org
+      .pull-right
+        .bullet
+      .pull-right
+        | 212.380.4460
+      .pull-right
+        .bullet
+      .pull-right
+        | 460 West 34th Street 17th Floor, New York, NY 10001-2320
```
jgoley: This can probably be handled by adding each contact method to a list item then changing the display property of the items to `inline-block`

imports/stylesheets/footer.import.styl
```diff
@@ -6,17 +6,29 @@ footer
...
     img
       margin-bottom 15px
+    a
+      bg('/images/eha-logo', 280px, 78px, 'svg')
+      display block
+      margin 10px auto
+  .contact
+    font-weight bold
+    margin-top 40px
+    color darkslategray
```
jgoley: Let's use the [color variables](https://github.com/ecohealthalliance/eidr-connect/blob/master/imports/stylesheets/variables.import.styl) whenever possible. Looks like these should be `$primary`

imports/stylesheets/footer.import.styl
```diff
@@ -6,17 +6,29 @@ footer
...
+    margin-right 2%
+  .bullet::before
+    padding 20px
+    content "\f111"
+    font-family FontAwesome
```
jgoley: I think this css can be avoided by using the fontawesome html classes: `.fa.fa-circle`

treyyoder:
@jgoley : Revised per your suggestions.

jgoley:
@treyyoder Thanks for making those changes! I added some responsive styles.
👍  @nathanathan I think this is good to go

PR #88, opened 2016-09-14
Updates the Curator Inbox details panel

gbersani:
Updates the right panel in the curator inbox view to match the most recent mock-ups.

Story: https://www.pivotaltracker.com/n/projects/1424346

Notes:
- Order of fields in the incidents table changed slightly form the mock-ups to keep it consistent with the table in the user-event view. 
- There is a bug with react-table package which I have not been able to find a work-around. Pagination/Sorting make the package confused when we have unrelated fields with the same key, causing sorting on child tables to affect the table above it.

Blocked:
- Suggested events should be filtered by metadata included in the collection created on https://www.pivotaltracker.com/story/show/129926009


nathanathan:
👍 

I don't see where the reactive-table bug is manifested, but it sounds like it might be possible to fix by calling event.stopPropagation on some of the events from the child table.

PR #90, opened 2016-09-15
Event details edit modal

jgoley:
Moves the edit event UI elements into a modal and changes the _admin options_ dropdown to a simple button near the title.

Event selection now is accomplished via a button in the modal. The alert confirmation is replaced with a secondary modal 'page' which asks the user for confirmation.

When there is no disease or summary, the user sees button which opens modal and the appropriate input field is focused.

Cleans up some jade/css - specifically switching from camelCase classnames to hyphenated and avoiding using the double curly brackets to reference properties.

PT Story: https://www.pivotaltracker.com/story/show/130376327

### DIFF
--- a/client/controllers/editEventModal.coffee
+++ b/client/controllers/editEventModal.coffee
@@ -1,0 +1,51 @@
+Template.editEventDetailsModal.onCreated ->
+  @confirmingDeletion = new ReactiveVar false
+
+Template.editEventDetailsModal.onRendered ->
+  instance = @
+  @$('#edit-event-modal').on 'show.bs.modal', (event) ->
+    instance.confirmingDeletion.set false
+    fieldToEdit = $(event.relatedTarget).data('editing')
+    # Wait for the the modal to open
+    # then focus input based on which edit button the user clicks
+    Meteor.setTimeout () ->
+      field = switch fieldToEdit
+        when 'disease' then 'input[name=eventDisease]'
+        when 'summary' then 'textarea'
+        else 'input:first'
+      instance.$(field).focus()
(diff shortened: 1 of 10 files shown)


client/controllers/editEventModal.coffee
```diff
@@ -0,0 +1,51 @@
...
+  'click .back-to-editing': (event, instance) ->
+    instance.confirmingDeletion.set false
+
+  'click .delete-event-confirmation': (event, instance) ->
+    Meteor.call 'deleteUserEvent', @_id, (error, result) ->
+      if not error
+        instance.$('#edit-event-modal').modal('hide')
+        $('.modal-backdrop').remove()
+        $('body').removeClass 'modal-open'
+        Router.go 'user-events'
```
jgoley: For some reason the modal backdrop is hanging around with just instance.$('#edit-event-modal').modal('hide'). I'm not sure whats going on so I remove the modal related elements/classes manually.

nathanathan: Using `Modal.hide(instance)` might fix it.

jgoley: That doesn't seem to work either. I also tried a timeout to give the dismissal time but that doesn't work so I'm stumped. Maybe @sturmer-eha has ideas?

jgoley: @nathanathan I just showed this to Yuri. We've concluded its a bug somewhere with bootstrap or Blaze making changes to the DOM such that bootstrap cannot find the backdrop.

nathanathan:
👍

PR #91, opened 2016-09-15
EIDR-C article pub dates can be edited

mawaruDOLL:
Since the database stores the publication date in the local machine's timezone, we need to convert it to the target timezone when showing it on the edit form. I tried to use moment timezone to handle this conversion but it messed up the date time picker plugin, which relies on moment.  I also noticed that if a time isn't selected when saving, the edit form will show the time as 12:00 AM the next time it's opened. Maybe the date and time should be stored separately in the database.

### DIFF
--- a/client/controllers/articles.coffee
+++ b/client/controllers/articles.coffee
@@ -25,11 +25,16 @@ Template.articles.helpers
           return ""
       }
     ]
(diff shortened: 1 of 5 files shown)


nathanathan:
👍

PR #92, opened 2016-09-15
Add button for adding non-suggested incident reports on incident report suggestion modal

mawaruDOLL:
The new button is in the modal footer and simply opens the new incident form modal.


nathanathan:
👍

PR #93, opened 2016-09-15
Add close buttons at the top of all modals

mawaruDOLL:


client/views/modalCloseButton.jade
```diff
@@ -0,0 +1,4 @@
+template(name="modalCloseButton")
+  button.close(type="button" data-dismiss="modal" aria-label="Close")
+    span(aria-hidden="true")
+      i.fa.fa-times
```
jgoley: This may have been picked-up too soon. In #90 I introduced an element and class ([here](https://github.com/ecohealthalliance/eidr-connect/blob/cf73099c699554bbbfc4fe16b626aa01178e4a61/client/views/userEvent.jade#L137)). Maybe once that's merged we can combine these ideas?

imports/stylesheets/event.import.styl
```diff
@@ -324,3 +324,6 @@ td.col-md-6.countCell.text-center
       li
         color #95989A
         padding 4px 0
+
+.modal-header h4.modal-title
+  line-height 2
```
jgoley: In #90 I introduced an independent file that contains general styles pertaining to modals. I think this change me be unnecessary after those changes.

jgoley:
@nathanathan I think we can still use this, it just needs to be modified/updated to match the patterns I introduced in #90

nathanathan:
👍

PR #95, opened 2016-09-15
add functionality to retrieve suggested event source articles

sturmer-eha:


client/controllers/articles.coffee
```diff
@@ -48,12 +48,12 @@ Template.articles.helpers
   selectedSource: ->
     selectedId = Template.instance().selectedSourceId.get()
     if selectedId
-      Articles.findOne selectedId
+      grid.Articles.findOne selectedId
```
nathanathan: It would be better not to use the grid object. We have a plans to get rid of it: https://www.pivotaltracker.com/story/show/129994355

server/interfaces.coffee
```diff
@@ -33,3 +34,15 @@ Meteor.methods
         dateUTC = match[1].replace(' ', 'T') + offset
         dateUTC
     )
+
+  queryForSuggestedArticles: (eventId) ->
+    event = grid.UserEvents.findOne(eventId)
+    console.log "Calling SPA API @ " + SPA_API_URL
+    response = HTTP.call('GET', "#{SPA_API_URL}/relatedArticles", {
+      params: { eventName: event.summary }
```
nathanathan: I would like the API for this to be more generic. We could use a "/find" endpoint that takes arbitrary mongo queries as a parameter. That way all the EIDR specific code for building the query from an EIDR event would stay in the EIDR codebase.

server/interfaces.coffee
```diff
@@ -1,6 +1,7 @@
 DateRegEx = /<span class="blue">Published Date:<\/span> ([^<]+)/
 
 GRITS_API_URL = process.env.GRITS_API_URL or "https://grits.eha.io/api/v1"
+SPA_API_URL = process.env.SPA_API_URL or "http://localhost:5940/api/v1"
```
nathanathan: Before we merge this, the localhost URL should be replaced by a spa.eha.io one.

server/interfaces.coffee
sturmer-eha: My only concern with that approach would be the ease of overloading the database engine by supplying a very permissive query. If we are okay to allow any queries to go through this API and intend to limit the access by whitelisting IPs, I'm totally alright with that.

nathanathan: Any queries are okay as long as they are find queries.

sturmer-eha:
This is basic but should be good for an initial implementation.
To test locally: `SPA_API_URL=http://localhost:5940/api/v1 make run`
Relies on locally running branch `external-api` of SPA.

client/controllers/articles.coffee
sturmer-eha: Done

server/interfaces.coffee
```diff
@@ -32,4 +35,25 @@ Meteor.methods
         offset = UTCOffsets[tz]
         dateUTC = match[1].replace(' ', 'T') + offset
         dateUTC
-    )
+
+  queryForSuggestedArticles: (eventId) ->
```
nathanathan: Could you make this remove suggestions that have already been added to the event?

server/interfaces.coffee
```diff
@@ -32,4 +35,25 @@ Meteor.methods
...
+    check eventId, String
+    event = grid.UserEvents.findOne(eventId)
+    console.log "Calling SPA API @ " + SPA_API_URL
+    unless event
+      throw new Meteor.Error 404, "Unable to fetch the requested event record"
+    keywords = _.uniq event.eventName.match(/\w{3,}/g)
```
nathanathan: The event disease should also be used to form the query.

client/views/sourceModal.jade
```diff
@@ -18,6 +18,14 @@ template(name="sourceModal")
...
                 else
                   input#article.form-control.new-article(type="url" name="article" required)
+            if suggestedArticles
+              .form-group
+                .col-md-12
+                  label.control-label Suggested Articles
+                  ul#suggested-articles
+                    each suggestedArticles
+                      li
+                        a(href=url) {{subject}}
```
nathanathan: These links should use a blue font

server/interfaces.coffee
```diff
@@ -32,4 +35,25 @@ Meteor.methods
...
+    mongoQuery = { $and: [] }
+    for keyword in keywords
```
nathanathan: I added a text search endpoint since it will be faster and allow sorting by text similarity: https://github.com/ecohealthalliance/spa/blob/master/server/api.coffee#L4
Could you use that instead? (The /find endpoint will be useful for other things like pulling new articles into the curator inbox.)

sturmer-eha:
I had to extend the /search API end-point to make the $nin part work for `promedId`

nathanathan:
👍 Looks good

PR #97, opened 2016-09-16
User can filter map using a date range picker

mawaruDOLL:
I was able to get filtering the event map by date ranges working. However, the date range picker plugin really has problems with inline styling on this page. So this branch has it attached to a text input as a start. Maybe someone else will be able to get it working inline or find a different library that has an inline option along with all of the other necessary features.

### DIFF
--- a/client/controllers/eventMap.coffee
+++ b/client/controllers/eventMap.coffee
@@ -77,7 +77,7 @@ Template.eventMap.onRendered ->
           eventIndex += 1
 
     instance.templateEvents.set templateEvents
-    instance.disablePrev.set if eventIndex < totalEventCount then false else true
+    instance.disablePrev.set if eventIndex < filteredEvents?.length then false else true
     instance.disableNext.set if currentPage is 0 then true else false
     if instance.allMapMarkers
       map.removeLayer instance.allMapMarkers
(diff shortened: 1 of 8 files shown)


jgoley:
@nathanathan This is ready for review.
I've changed the calendars to inline and optimized the styling for different screen resolutions.
I also spent a fair amount of time refining the UX.

client/controllers/dateSelector.coffee
```diff
@@ -0,0 +1,67 @@
+createInlineDateRangePicker = require '/imports/ui/inlineDateRangePicker'
+{ setVariables,
+  clearDateRange }          = require '/imports/ui/setRange'
```
nathanathan: @jgoley It looks like the setRange file is missing.

jgoley: @nathanathan My bad, I changed the file name just before pushing and forgot to add the file. Sorry about that.

jgoley:
@nathanathan I've removed the before/after filtering options as you requested in Slack. 
I also added an _Apply_ button which solves the bug with manual entry not triggering an update.

nathanathan:
👍 Looks good!

PR #98, opened 2016-09-20
Adds range filtering for the curator inbox

gbersani:
Story:
https://www.pivotaltracker.com/story/show/129926215

- Adds a calendar to the Curator Inbox
- Adds server-side filter for the subscription
- Makes many UI updates to the curator inbox to allow range filtering using multiple sectioned reactive-tables


nathanathan:
👍 Looks good!

PR #99, opened 2016-09-21
Import nib

jgoley:
Uses Nib which is included in the Meteor Stylus package.
Also removes browser specific properties that will now be added by Nib.


nathanathan:
👍

PR #100, opened 2016-09-21
Replace explicit media queries with rupture mixins

jgoley:
Remove unused styles and move like styles to independent files

This breaks some of the mobile styles...I think we should probably fix those separately...this PR is already too big.


nathanathan:
@jgoley Merging the last couple PRs created conflicts on this one.

nathanathan:
👍

PR #101, opened 2016-09-22
Adds article acceptance and an additional filter

gbersani:
PR takes care of 2 related stories:
https://www.pivotaltracker.com/story/show/129926131
https://www.pivotaltracker.com/story/show/130742703

- Adds option to mark articles as reviewed.
- Adds additional “reviewed” filter for the curator inbox.
- Fixes minor bug with the tables on the inbox not displaying all articles.

There appears to be some odd behavior caused by us having multiple redundant/duplicate articles in the collection for our testing. For instance, accepting an article might shift it’s order on the list.

This should be remedied when implementing the story: https://www.pivotaltracker.com/story/show/129926009

### DIFF
--- a/client/controllers/curatorInbox.coffee
+++ b/client/controllers/curatorInbox.coffee
@@ -182,6 +182,7 @@ Template.curatorInboxSection.helpers
       fields: fields
       showRowCount: false
       showFilter: false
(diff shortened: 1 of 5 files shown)


nathanathan:
👍

PR #107, opened 2016-10-05
Detect whether counts are cumulative for suggested incidents

nathanathan:


client/controllers/suggestedIncidents.coffee
```diff
@@ -148,10 +148,18 @@ Template.suggestedIncidentsModal.onCreated ->
...
         incident.countAnnotation = keypoint
+        countValue = keypoint.count.number or keypoint.count.range_end
         if keypoint.count.case
-          incident.cases = keypoint.count.number
+          incident.cases = countValue
         else if keypoint.count.death
-          incident.deaths = keypoint.count.number
+          incident.deaths = countValue
+        # Detect whether count is cumulative
+        if "incremental" of keypoint.count
```
Broham: I haven't really seen `of` used like this before.  Isn't it typically used for looping over object properties?  Could you explain what is going on here just so I have a clearer picture on my end?

PR #111, opened 2016-10-12
Incident scatter plot

dan-nyanko:
This PR adds modular/class-based D3.js scatter plot.

The base plot class has the following features:
  - x, y grid lines
  - x, y axes 
    - custom labels
    - supports types of 'numeric' and 'datetime' 
  - zoomable area
  - tooltips

The scatter plot extends the base plot to display custom rectangular markers.

The modules can be easily extended for additional plot types, such as, line or bar.  The modules are contained within `imports/charts` and are as follows:

```
Axes.coffee
Grid.coffee
Plot.coffee
RectMarker.coffee
ScatterPlot.coffee
Tooltip.coffee
Zoom.coffee
```

In addition, we could turn this into a standalone library that is published on npm and/or atmosphere.

An example of building the scatter plot:

```
    # build the plot
    @plot = new ScatterPlot({
      containerID: 'scatterPlot',
      svgContainerClass: 'scatterPlot-container',
      height: $('#event-incidents-table').parent().height(),
      axes: {
        # show grid lines
        grid: true,
        x: {
          title: 'Time',
          type: 'datetime',
          minMax: [
            Axes.minDatetime(_.pluck(data, 'x')),
            Axes.maxDatetime(_.pluck(data, 'w')),
          ],
        },
        y: {
          title: 'Incidents',
          type: 'numeric',
          minMax: [
            0,
            Axes.maxNumeric(_.pluck(data, 'y')),
          ]
        }
      },
      tooltip: {
        opacity: .8
        # function to render the tooltip
        template: (marker) ->
          marker.moment = moment # template reference for momentjs
          marker.incidents = 'incidents'
          if marker.y <= 1
            marker.incidents = 'incident'
          # underscore compiled template
          tmpl = _.template(tooltipTmpl)
          # render the template
          tmpl(marker)
      },
      # enable plot zooming
      zoom: true,
    })
```


nathanathan:
This is really cool! I like the idea of publishing it as as a standalone somewhere. There are a few minor issues I noticed:
 * I upgraded the version of coffeescript because an error was preventing the page from loading with the previous version.
 * The markers all appear to be 1.5 days long. I think this is caused by the time in the date-range start being truncated but not in the date-range end.
 * The bottom boarder only appears when zoomed.
 * When a new incident report is added, the plot doesn't update.

dan-nyanko:
@nathannathan - Please elaborate on the bottom border appears when zoomed?  What browser and do you have a screenshot?

dan-nyanko:
@nathanathan - Also, what are the steps to adding an incident?  So, for example, if the user navigates backwards, fills out and submits a form, when they return to the 'Incidents Report' tab the data is missing from the incidents array?  I assume a page refresh would make it appear?  Or are you expecting the plot to be reactive?  Thanks.

nathanathan:
Here's a screenshot of the missing boarder under the default zoom level:
screen shot 2016-10-13 at 3 34 04 pm

I'm using Chrome 53.0.2785.116 on a mac.

If there isn't a quick fix for it, I think it's alright as is.

dan-nyanko:
@nathanathan ping

nathanathan:
👍 Looks great! The bottom boarder is fixed.

PR #116, opened 2016-10-18
bugfix - debounce plot.update; the plot will determine if it should show/hide the no data warning

dan-nyanko:
bugfix - https://www.pivotaltracker.com/story/show/132659361

- The plot will determine if it should show/hide the no data warning.
- improved styling
- improved center positioning 
- added transition on the opacity of the showWarn message
- misc bugfix to disable resetZoom if data is not at least one or zoom is disabled
- added transition on the opacity to the first drawing of the plot

Steps A:
1. goto the Event Map
2. click on a marker
3. click on a link to /user-event/:id
4. confirm that the warning does not appear

Steps B:
1. goto the Tracked Events
2. sort by Article Count ascending
3. click on an event to /user-event/:id
4. confirm that the warning appears for an event without data


nathanathan:
👍

PR #119, opened 2016-10-20
adds filter capability to the plot and cumultive filter button

dan-nyanko:
https://www.pivotaltracker.com/story/show/132223713

This feature added an API to add/remove filters to the base plot. This allows the plot instance to apply these filters on each draw, including the default filter for the current axes domain. This simplifies a lot of complexity that may otherwise occur in the controller code trying to synchronize data state between each plot update. 

In the following example, the filtering method is given a name and method to execute. Note: the method will be bound to the plot instance.

```
addFilter('cumulative', (d) ->
  // do something with a data element
```

```
removeFilter('cumulative')
```

image

### DIFF
--- a/client/controllers/incidentReports.coffee
+++ b/client/controllers/incidentReports.coffee
@@ -74,6 +74,7 @@ Template.incidentReports.onRendered ->
       if incident
         return incident
     )
+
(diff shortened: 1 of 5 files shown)


nathanathan:
Looks good :+1:

PR #120, opened 2016-10-20
reverses pagination so UI doesn't appear to start at the end

dan-nyanko:
https://www.pivotaltracker.com/story/show/130367327


nathanathan:
👍

PR #123, opened 2016-10-21
Fix publication date autofill bug

nathanathan:
Details about the bug: https://www.pivotaltracker.com/story/show/130514787

This also fixes a bug in eliminating previously added articles from the article suggestions.


Broham:
Looks good :+1:

PR #125, opened 2016-10-24
Move reviewed button and undo hover state

jgoley:
Moves the review (formally _accept_) button and combine toggling into single button.


nathanathan:
👍

PR #128, opened 2016-10-25
add feature to suggest user events based on title of promed post

dan-nyanko:
https://www.pivotaltracker.com/story/show/132240647

- $text index on UserEvent `eventName` and `summary`
- server-side meteor method to perform the $text $search sorted by score
- reactive table that displays the suggested events 
- updates reactively based on associatedEventIdsToArticles object

Steps:
1. Visit Curators -> Inbox
2. Note: collapsible 'Suggested Events' table, matches that do not exist in 'Associated Events' will be displayed
3. Click on the `+` icon to add an event to the 'Associated Events' section
4. Click on the `-` icon to remove and event from the 'Associated Events' section
5. Note: the table updates reactively

### DIFF
--- a/client/controllers/curatorEvents.coffee
+++ b/client/controllers/curatorEvents.coffee
@@ -16,7 +16,6 @@ Template.curatorEvents.onCreated ->
     ).map((article)->
       [article.userEventId, article]
     ))
-  Template.curatorEvents.associatedEventIdsToArticles = @associatedEventIdsToArticles
 
 Template.curatorEvents.onRendered ->
   @$('#curatorEventsFilter input').attr 'placeholder', 'Search events'
@@ -33,6 +32,12 @@ Template.curatorEvents.helpers
       _id:
         $in: _.keys(Template.instance().associatedEventIdsToArticles.get())
     )
+
+  associatedEventIdsToArticles: ->
+    Template.instance().associatedEventIdsToArticles
+
+  title: ->
+    Template.instance().data.title
 
   associated: () ->
     articleId = Template.instance().data._id
--- a/client/controllers/curatorEvents.coffee
+++ b/client/controllers/curatorEvents.coffee
@@ -16,6 +16,7 @@ Template.curatorEvents.onCreated ->
     ).map((article)->
       [article.userEventId, article]
     ))
+  Template.curatorEvents.associatedEventIdsToArticles = @associatedEventIdsToArticles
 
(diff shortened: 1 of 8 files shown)


methods/userEvents.coffee
```diff
@@ -61,3 +61,8 @@ Meteor.methods
         $unset:
           lastIncidentDate: ""
       })
+
+if Meteor.isServer
```
nathanathan: I think this would be a better fit for server/interfaces.coffee since it is a server only method.

client/views/curatorSuggestedEvents.jade
```diff
@@ -0,0 +1,11 @@
...
+  .curator-events-header.curator-collapse
+    h4
+      | Suggested Events
+    .curator-events-collapse
+      if isOpen
+        i.fa.fa-lg.fa-chevron-circle-down
+      else
+        i.fa.fa-lg.fa-chevron-circle-left
+  if isOpen
+    +reactiveTable collection=events settings=settings class='table curator-events-table'}
```
nathanathan: The `}` at the end looks out of place

client/controllers/curatorEvents.coffee
```diff
@@ -16,6 +16,7 @@ Template.curatorEvents.onCreated ->
     ).map((article)->
       [article.userEventId, article]
     ))
+  Template.curatorEvents.associatedEventIdsToArticles = @associatedEventIdsToArticles
```
dan-nyanko: Please provide an example of passing a reactive var to another template through a helper.

methods/userEvents.coffee
dan-nyanko: It seemed link an intuitive place to put all the meteor methods.  So this folder is only for `dual-purpose` code, e.g both client and server?

client/controllers/curatorEvents.coffee
nathanathan: Here is an example for the reviewFilter variable:
https://github.com/ecohealthalliance/eidr-connect/blob/master/client/views/curatorInbox.jade#L28

methods/userEvents.coffee
nathanathan: Ideally, yes.

dan-nyanko: Ok.

nathanathan:
👍

PR #129, opened 2016-10-25
fixing multiple click bug

Broham:
Fixes a bug where the first record in the inbox was being clicked multiple times


nathanathan:
👍

PR #130, opened 2016-10-25
changing threshold at which we show H:M for x axis

Broham:
This should resolve the issue of duplicate labels showing at certain zoom levels.


nathanathan:
👍

PR #131, opened 2016-10-25
adds feature to clean urls

dan-nyanko:
https://www.pivotaltracker.com/story/show/131123351

## Server

- adds startup methods to clean urls from the substrings `https?` and `(www.)?`
Example: http://www.promedmail.org/post/abc -> promedmail.org/post/abc

Steps:
1. Drop all `articles`, `counts`, and `curatorSources` collections from mongo
2. Perform mongorestore from backup that contains urls/links that need cleaned
2a. Extract a backup from AWS (this will create the directory `/var/log/dump/eidr-connect`)
2b. Change directories `cd /var/log/dump/eidr-connect`
2c. Run mongorestore `mongorestore -d eidr-connect .`
3. Verify that urls contain `https?` and `(www.)?`
4. Start meteor server
4a. Watch for output to stdout about bulkUpdate nModified for each collection
4b. Notice `curatorSources` falls back to single item updates due to its use of Mongo.ObjectID
5. Verify that the urls/metadata.links in `articles`, `counts`, and `curatorSources` appear correct.
6. Restart meteor server; Note: the queries return 0

## Client
Note: The client displays `formatUrl()` with http:// reinserted
image
image

## Notes
- @nathanathan - I started introducing CoffeeScript `import` and `export` methods vs. `require`. This allows for a slightly cleaner interface when importing individual methods. See CoffeeScript [ES6 Style Modules](http://coffeescript.org/#modules).  I can convert all the modules to use this syntax or revert.

- The backup `mongodump-2016-10-25.tar` has data that should be manually cleaned if its still on production.  See the following query: `db.getCollection('curatorSources').find({'_id': {$in: [ObjectId('57f6bb9bffd3ba063abcef57'), ObjectId('57ae633f04594a0e1adb626a')]}})`

- The implementation of this feature does introduce a slight bit of complexity. E.g. within the client it will be necessary to call `utils.cleanUrl()` prior to saving to the database (strips https? and www) and then `utils.formatUrl()` to reinsert the `http://` so that `<a>` functions properly within templates.

### DIFF
--- a/server/cleanUrls.coffee
+++ b/server/cleanUrls.coffee
@@ -1,0 +1,96 @@
+import Articles from '/imports/collections/articles.coffee'
+import IncidentReports from '/imports/collections/incidentReports.coffee'
+import CuratorSources from '/imports/collections/curatorSources.coffee'
+import { attemptBulkUpdate } from './dbUtils.coffee'
+import { cleanUrl } from '/imports/utils.coffee'
+
+###
+# updateUrls - looks for `https?` and `(www.)` patterns within the url
+#   property of the specified collection. @note this requires that the document
+#   has a `url` property as either a string or an array
+#
+# @param {object} collection, the mongo collection to search aggregately
+# @param {string} host, the hostname of the url to clean
+# @param {string} domain, the domain of the url to clean
+###
+updateUrls = (collection, host, domain) ->
+  bulkUpdates = []
+  parts = [new RegExp('^(?:https?:\/\/)', 'i'), new RegExp("^(www\.)#{host}\.#{domain}", 'i')]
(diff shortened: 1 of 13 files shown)


server/cleanUrls.coffee
```diff
@@ -0,0 +1,96 @@
...
+#   `CuratorSources` urls/links properties.
+###
+cleanPromedUrls = ->
+    host = 'promedmail'
+    domain = 'org'
+    updateUrls(Articles, host, domain)
+    updateUrls(IncidentReports, host, domain)
+    updateMetadataLinks(CuratorSources, host, domain)
+
+if Meteor.isServer
```
nathanathan: The ifServer check is unnecessary for files in the server directory.

nathanathan:
👍 I like the import/export syntax. Let's convert all the require calls to it in another PR.

PR #132, opened 2016-10-26
Fix article publication date issues

nathanathan:
Changes:
* Fixes bug where use of isDST function only detects DST if the client's locale has DST.
* Rectifies publication dates and timezones using pub dates from the ProMED posts collection.
* Adds a schema for articles collection which is used to validate them on start-up.


Broham:
looks good :+1:

PR #133, opened 2016-10-26
Show next source after reviewal

jgoley:
This selects the next unreviewed source when a curator marks the currently selected source as reviewed.

Also reworks how we display source details in the curator inbox:
- Uses template in jade instead of calling `Blaze.renderWithData` in the controller
- Creates a reactiveVar which tracks which source is selected (used in the reactive table and the details template)
- Notifies user that next source is being prepared or retrieved after reviewal

And cleans up the styles of the main source table

I noticed a possible issue. [This method](https://github.com/ecohealthalliance/eidr-connect/blob/master/methods/promedPosts.coffee#L29-L41) seems to overwrite some properties such as `addedDate` locally? I wanted to find a place to add a `reviewed` property and set it false to prevent having to run [this sort of query](https://github.com/ecohealthalliance/eidr-connect/blob/39e46fb540480f4912cb57d020ef1845dfecc4e3/client/controllers/curatorInbox.coffee#L42-L46). But if I added it to the method mentioned above, it is reset every time the view is loaded. Maybe I'm doing this in the wrong place?

### DIFF
--- a/client/controllers/curatorEvents.coffee
+++ b/client/controllers/curatorEvents.coffee
@@ -4,15 +4,20 @@ CuratorSources = require '/imports/collections/curatorSources.coffee'
 Articles = require '/imports/collections/articles.coffee'
 
 Template.curatorEvents.onCreated ->
-  @subscribe("articles", {
-    url:
-      $regex: "post\/" + @data._sourceId + "$"
-  })
-  @associatedEventIdsToArticles = new ReactiveVar([])
+  instance = @
+  @autorun ->
+    sourceId = instance.data.selectedSourceId.get()
+
+    instance.subscribe("articles", {
+      url:
+        $regex: "post\/" + sourceId + "$"
+    })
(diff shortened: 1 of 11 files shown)


nathanathan:
Maybe you could use $setOnInsert for the "reviewed" property to keep it from being overwritten when the collection is updated.

jgoley:
I was looking for something that had this sort of functionality. I'll update for `reviewed` and `addedDate`

jgoley:
@nathanathan I've updated the last commit to use `$setOnInsert`. Looks like, though, we will need to run a script to update the existing documents in the `CuratorSources` collection — creating a `reviewed` property if one does not exist?

nathanathan:
Yeah, could you add a Meteor.startup function that does that?

jgoley:
Just updated the startup script to add the `reviewed` property if it does not exist.

nathanathan:
👍 Looks great!

PR #134, opened 2016-10-26
adding 'X of Y incidents reviewed' text to the Suggested Incident Rep…

Broham:
…orts modal

### DIFF
--- a/client/controllers/suggestedIncidents.coffee
+++ b/client/controllers/suggestedIncidents.coffee
@@ -1,4 +1,5 @@
 incidentReportSchema = require('/imports/schemas/incidentReport.coffee')
+Incidents = require '/imports/collections/incidentReports.coffee'
 # A annotation's territory is the sentence containing it,
 # and all the following sentences until the next annotation.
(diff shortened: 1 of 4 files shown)


client/controllers/suggestedIncidents.coffee
```diff
@@ -172,11 +174,18 @@ Template.suggestedIncidentsModal.helpers
     Template.instance().incidentCollection.find().count() > 0
   loading: ->
     Template.instance().loading.get()
+  annotatedCount: ->
+    if Template.instance().annotationCount.get()
+      count = Incidents.find({userEventId: Template.instance().data.userEventId}).fetch().length
```
Broham: I might be a bit confused on terminology, but I think that it is the number of incidents.  The `count` variable is the number of records in the `counts` collection (referred to here as `Incidents`).  The number of suggested incidents is the `annotationCount` reactive var.  Does this sound correct, or am I counting the wrong things?

nathanathan:
👍

PR #137, opened 2016-10-26
Adding title to doc when adding source.  

Broham:
If a title is available for the document we will now display it.  Will have a separate job that will update documents to include the title.

### DIFF
--- a/client/controllers/articles.coffee
+++ b/client/controllers/articles.coffee
@@ -11,7 +11,8 @@ Template.articles.helpers
         key: "url"
         label: "Title"
         fn: (value, object, key) ->
-          return value
+          # switching over to displaying the title in this column.  If that's not loaded in the DB show the URL.
+          return object.title || value
       },
       {
         key: "addedDate"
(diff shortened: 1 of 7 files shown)


nathanathan:
This should also provide a way to set the title for articles that are added by url rather than from the suggestions, save the title of articles added through the curator inbox, and update the article schema.

Broham:
ok @nathanathan I pushed an update that includes a startup script to set the promed titles.  Take a look and let me know if this is what you had in mind.

nathanathan:
@Broham The update looks great. That is what I had in mind.

Broham:
@nathanathan is there anything else I need to do on this guy?

nathanathan:
@Broham Yes, you should also provide a way for users to set the title for articles that are added by url and set the title of articles added through the curator inbox.

Broham:
This should be good to go now @nathanathan

nathanathan:
The title field on the article form is not being saved.

Broham:
@nathanathan fixed

nathanathan:
👍 Looks good

PR #141, opened 2016-10-28
fix missing title property

dan-nyanko:
Fixes the missing title prop for the curatorSuggestedEvents template by query against curatorSources using the new reactive var `selectedSourceId`

### DIFF
--- a/client/controllers/curatorSuggestedEvents.coffee
+++ b/client/controllers/curatorSuggestedEvents.coffee
@@ -48,7 +48,7 @@ Template.suggestedEvents.onCreated ->
     @events.set(events)
 
   @autorun =>
-    source = CuratorSources.find({_id: @data.selectedSourceId.get()}).fetch()[0]
+    source = CuratorSources.findOne({_id: @data.selectedSourceId.get()})
     if source
       # strip punctuation and create tokens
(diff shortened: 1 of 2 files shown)


client/controllers/curatorSuggestedEvents.coffee
```diff
@@ -46,20 +47,23 @@ Template.suggestedEvents.onCreated ->
...
     @events.set(events)
 
-  # strip punctuation and create tokens
-  tokens = if @data.title then @data.title.replace(/[~`!@#$%^&*(){}\[\];:"'<,.>?\/\\|_+=-]/g, '').split(' ') else []
-  # remove the first token from the search string (PRO/AH/EDR> without the special chars)
-  # TODO: this is domain specific to ProMed. If other feeds are added, we would
-  # need to do some conditional logic `if feed.isPromed then remove first token`.
-  search = if tokens.length > 1 then tokens.slice(1, tokens.length).join(' ') else ''
+  @autorun =>
+    source = CuratorSources.find({_id: @data.selectedSourceId.get()}).fetch()[0]
```
Broham: It might be cleaner to use `findOne()` here instead of `find().fetch()[0]`

PR #143, opened 2016-10-28
Do not show date header if there are no corresponding sources

jgoley:
PT: https://www.pivotaltracker.com/story/show/133191391

Also fixes a bug with searching the source list: https://www.pivotaltracker.com/story/show/133359851


nathanathan:
👍

PR #144, opened 2016-10-31
Refactor templates and polish style

jgoley:
Starts to rework the UI code and styles of the event page


nathanathan:
The event sources panel has a spacing issue:
screen shot 2016-11-01 at 12 39 25 pm

jgoley:
My bad. I was going to work on that in another PR and didn't realize the spacing was weird.
I've pushed a commit which fixes the issue styles that area.

nathanathan:
👍 Looks good

PR #145, opened 2016-11-01
scatterPlot markers as line segments

dan-nyanko:
# Feature to add scatterPlot markers as line segments

- upgraded to d3v4 via `npm`

Updates to the ScatterPlot interface:
- created concept of a Group which contains Nodes
- each node should be an instanceof `Node`
- each group should be an instanceof `Group`
- throws errors `InvalidNodeError` and `InvalidGroupError`
- `plot.update()` accepts an array or an object
  - if its an array: a default group will be created and all nodes will be appended
  - if its an object: each key will be a new group and its value as an array of nodes
- the creation of nodes and groups is transparent to keep the API simple
- when constructing the ScatterPlot, there is an option for `group`, which allows binding an event handler during onEnter.
```
    # group events
    group: {
      # methods to be applied when a new group is created
      onEnter: () ->
        # bind mouseover/mouseout events to the plot tooltip
        @.group.on('mouseover', () =>
          if @plot.tooltip
            @plot.tooltip.mouseover(@, d3.event.pageX, d3.event.pageY)
        ).on('mouseout', () =>
          if @plot.tooltip
            @plot.tooltip.mouseout()
        )
    },
```

New type of marker `SegmentMarker`:
- has start, end point with slope of 0
- `SegmentMarker.groupOverlappingSegments()` will group overlapping segments on the y-axis together, so the tooltip displays multiple overlapping nodes.

image
 
image

https://www.pivotaltracker.com/story/show/132224301

### DIFF
--- a/.meteor/packages
+++ b/.meteor/packages
@@ -9,7 +9,6 @@ coffeescript@1.11.1_1
 stylus@2.513.5_1
 iron:router
 twbs:bootstrap
-d3
 okgrow:analytics
 http@1.2.9_1
 fortawesome:fontawesome
@@ -51,4 +50,3 @@ shell-server@0.2.1
 aldeed:simple-schema
 dangrossman:bootstrap-daterangepicker
 stevezhu:lodash
-d3js:d3
--- a/.meteor/versions
+++ b/.meteor/versions
@@ -22,8 +22,6 @@ callback-hook@1.0.9
 check@1.2.3
 chrismbeckett:toastr@2.1.2_1
 coffeescript@1.11.1_1
-d3@1.0.0
-d3js:d3@3.5.8
 dangrossman:bootstrap-daterangepicker@2.1.13
 ddp@1.2.5
 ddp-client@1.3.1_1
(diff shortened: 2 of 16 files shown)


toph-allen:
Interesting. It looks good! I have one thought:

I think this is good for now, and think we should go ahead and merge, but I think that as we hone our visualizations, we're going to want [an option for] the Y axis to display **count / period** — in other words, the Y axis would then display the relative number of cases expected over X period of time. I think that aligns more with the way that epidemiologists think about this stuff, and probably is a better visual representation for what the magnitude of the number actually represents.

toph-allen:
Dammit, I typed a comment but the page crashed when I submitted it.

I think this looks good for now. In the future we'll want [an option for] the Y axis to display **count / period**. I think that has better visual fidelity to what's actually going on (i.e. a count of fifty over fifty days is as severe, given the same population, as a count of ten over ten days), and probably aligns better with how an infectious disease analyst would conceive of the problem.

nathanathan:
👍 Looks great!

PR #149, opened 2016-11-03
Update styles of headers and list items

jgoley:
This was a little independent of the chore I'm working on so I thought I'd open a PR specifically for it.


nathanathan:
👍

PR #150, opened 2016-11-04
Event page incident details styles

jgoley:
Updates the Incident Report details area like so:
screen shot 2016-11-04 at 3 53 26 pm

It looks like there's a bunch of changes but its mostly cleaning up code styling. Like using `'` instead of `"`, removing excessive curly brackets for object declarations, etc.
We probably need to establish common patterns for stuff like this and establish a JSLint file. 
I noticed we use 3 different naming conventions to refer to the template instance in events: `instance`, `template` and `templateInstance`. I think in recent projects we've used `instance`. Maybe we should fix all of these at some point?

(layout may break on smaller screens — I can update those styles in another PR)

PT Chore: https://www.pivotaltracker.com/story/show/133691167

### DIFF
--- a/client/controllers/articles.coffee
+++ b/client/controllers/articles.coffee
@@ -44,37 +44,28 @@ Template.articles.helpers
           value
       }
     ]
-    if Roles.userIsInRole(Meteor.userId(), ['admin'])
-      fields.push({
-        key: "delete"
-        label: ""
-        cellClass: "remove-row"
-      })
-      fields.push({
-        key: "Edit"
-        label: ""
-        cellClass: "edit-row"
-      })
-    fields.push({
+
+    fields.push
       key: "expand"
       label: ""
-      cellClass: "open-row-right"
-    })
-    {
-      id: 'event-sources-table'
-      fields: fields
-      showFilter: false
-      showNavigationRowsPerPage: false
-      showRowCount: false
-      class: "table"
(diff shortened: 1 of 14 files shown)


nathanathan:
Let's make the tater style guide our gold standard for common patterns https://github.com/ecohealthalliance/tater/wiki/Coffeescript-Style-Guide . `instance` sounds like a good standard variable name. I would recommend a gradual Locke's socks style of clean-up as we add features and fix bugs.

@jgoley @Broham @dan-nyanko @arnoldbm @frosario @toph-allen Does that sound good to everyone?

nathanathan:
I'm seeing a few bugs when editing incident reports on this branch. The status and incident type selectors are not being filled in and I get an error message when attempting to save changes.

jgoley:
@nathanathan 1e2e354 fixes the issues with form inputs not receiving values, saving the incident report, and issues with the incident reports table after the incident report is saved.
I spent some time trying to keep the details of the incident report open and visible after save but if the table is sorted and the user changes the value of the sorted property of the report, the reactivity of the table moves the report. We should probably fix this somehow in the future. Having both the details and the report disappear isn't a great user experience.
We may need to switch to opening a modal when the user clicks on an incident report for more details or open the details temporarily fill the right pane (like the event sources UI).

nathanathan:
👍 Looks good!

PR #152, opened 2016-11-07
Cull nav items and decrease height

jgoley:
Compress account management to a single view

This also fixes some of the bugs associated with the header at smaller screens but doesn't fix them all. I'll fix those later, or may implement a header like in NIAM.

Updates to the account management views are temporary and just a start.

PT Story: https://www.pivotaltracker.com/story/show/133456939


nathanathan:
👍

PR #154, opened 2016-11-09
Make incident report addition more explicit

jgoley:
This shows an _add incident_ button (along with the info button) inline.


nathanathan:
👍

PR #155, opened 2016-11-09
Curator inbox add new event

jgoley:
Allows users to add events in the curator inbox. When an event is added, its automatically associated with the selected source. It modifies the edit event modal — combines the adding and editing events.

PT Story: https://www.pivotaltracker.com/story/show/128147377
PT Bug: https://www.pivotaltracker.com/story/show/133516463

### DIFF
--- a/client/controllers/curatorEvents.coffee
+++ b/client/controllers/curatorEvents.coffee
@@ -118,7 +118,7 @@ Template.curatorEvents.events
   "click .add-new-event": (event, instance) ->
     Modal.show 'editEventDetailsModal',
       action: 'add'
-      saveActionMessage: 'Save & Add to Source'
+      saveActionMessage: 'Add Event & Associate with Source'
       addToSource: true
       sourceId: instance.data.selectedSourceId.get()
       eventName: $('#curatorEventsFilter input').val().trim() or ''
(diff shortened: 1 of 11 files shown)


nathanathan:
I'm getting a "NaN" for the source count when I create a new event. Also, I suggest changing the button text to "Create Event & Associate Source"

jgoley:
@nathanathan Is this happening in the UI or on the object? When you add from the inbox and/or the tracked events page?

nathanathan:
It happens in the UI when I add an event from the inbox.

jgoley:
@nathanathan I found the problem. 3e5ec5d updates event creation/editing—uses both `$set` and `$setOnInsert`. I also updated the button text of the add event modal.

nathanathan:
Creating new events from the Tracked Events page isn't working for me.

jgoley:
@nathanathan My bad...apparently my global search for the method name failed to find and replace the one in the click event.
PS: I think we need think about reorganizing the structure of views and controllers—create directories for files that relate and dedicate individual files to each template with consistent naming conventions. I find it confusing when we conglomerate logic of several templates into one file.

nathanathan:
👍

PR #156, opened 2016-11-09
Curator inbox header ui updates

jgoley:
I'm having trouble focusing the input after clicking the icon.
Maybe this animation isn't necessary?

PT Chore: https://www.pivotaltracker.com/story/show/133627883


nathanathan:
Looks great! I'll go ahead and merge this since the focusing isn't much of an issue, although if we have to choose between them, ease of use is preferable to animations.

PR #157, opened 2016-11-10
Update inbox tooltips - do not show after click

jgoley:
- Sets tooltip color
- Adjusts placement
- removes tooltip after click


nathanathan:
👍

PR #161, opened 2016-11-10
Show add report/source btns if none and content if exists

jgoley:
This may require some more work. I'm not sure the suggested articles logic is working since I'm calling the modal without delivering some data.


nathanathan:
The controls for editing event details are not working. Article suggestions appear to be working.

jgoley:
Accidentally removed the reference to the modal template, sorry about that. Should be good to go now.

nathanathan:
👍

PR #163, opened 2016-11-11
Refactor delete confirmation into one template

jgoley:
We had 3 instances of the delete confirmation template/logic so I combined it into two  templates/controllers: one for the modal and another for its body since the body content may need to be different (e.g. event editing/deletion)

Also adds back source deletion/editing (adds it to the source detail view not in the table row as in the past)

PT Bug: https://www.pivotaltracker.com/story/show/134131349


Broham:
Looks good :+1:

PR #164, opened 2016-11-14
Tracked events update ui

jgoley:
Updates the UI of tracked events.
Also switches to a modal for housing the create event UI instead of a dedicated route.
And adds a loader to the table (in a kind of hacky way — the table is just added to the UI in front of the loader and the loading animation lingers.

PT Chore: https://www.pivotaltracker.com/story/show/133613311


nathanathan:
Looks good, but I would like the event modification dates to remain accessible for sorting. That is useful for keeping track of changes.

jgoley:
Sure thing. 92f6756 adds back that column and fix a small visual issue with the add event button.

nathanathan:
👍

PR #165, opened 2016-11-14
Fix responsiveness of event table and ensure table content is visible

jgoley:
Related to [this bug](https://www.pivotaltracker.com/story/show/134287587).
It refactors the mobile table styles (changes the breakpoint at which tables change to the mobile view)
I also discovered we were referencing the incorrect stylus file in [this file](https://github.com/ecohealthalliance/eidr-connect/blob/master/client/main.styl) `index.styl` is where we need to add stylus partials going forward.


nathanathan:
The issued in the linked bug still appear to be present in this branch.

jgoley:
@nathanathan Do you mind sharing a screenshot to give a little more context?

nathanathan:
screen shot 2016-11-16 at 12 25 16 am

Note the missing case/death numbers and the missing second date on the date range in the final row.

jgoley:
@nathanathan cd02af6 fixes those issues. Sorry, I thought they were visual bugs. 
I removed the `pluralize` npm package since it was redundant and export the existing `pluralize` function instead of it being a global.

nathanathan:
👍

PR #166, opened 2016-11-15
fixing escape to close modals bug

Broham:
right now if there are two modals on a page the "escape to close" functionality will not work.


nathanathan:
👍

PR #169, opened 2016-11-16
Make search inputs consistent

jgoley:
Also fixes the focus issue with the initially hidden search input in the curator inbox left pane header.


nathanathan:
👍

PR #173, opened 2016-11-17
User can add a feed url to db

jgoley:
Sets up basic functionality for adding feed urls.
I wrote out tests locally but I'm having trouble running them locally. I'll wait until we have tests fully functional then push those up.

PT story: https://www.pivotaltracker.com/story/show/133029117


client/controllers/feeds.coffee
```diff
@@ -0,0 +1,19 @@
+Feeds = require '/imports/collections/feeds.coffee'
+
+Template.feeds.onCreated ->
+  Meteor.subscribe 'feeds'
```
nathanathan: Using `@subscribe` would be better so the subscription is stopped when the template is destroyed.

client/router.coffee
```diff
@@ -91,3 +91,10 @@ Router.route "/user-event/:_id/:_view?",
     userEvent: UserEvents.findOne({'_id': @params._id})
     articles: Articles.find({'userEventId': @params._id}, {sort: {publishDate: -1}}).fetch()
     incidents: Incidents.find({'userEventId': @params._id}, {sort: {date: -1}}).fetch()
+
+Router.route "/feeds",
+  onBeforeAction: ->
+    unless Roles.userIsInRole(Meteor.userId(), ['admin'])
+      @redirect '/'
```
nathanathan: If the user visits the feeds url directly, this could redirect them before the login process finishes.

methods/feeds.coffee
```diff
@@ -0,0 +1,11 @@
+Feeds = require '/imports/collections/feeds'
+
+Meteor.methods
+  addFeed: (feed) ->
```
nathanathan: We should check that the feed url is not a duplicate.

client/controllers/feeds.coffee
```diff
@@ -13,7 +13,13 @@ Template.feeds.events
     if !/^(f|ht)tps?:\/\//i.test(feedUrl)
       feedUrl = "http://#{feedUrl}"
 
+    if Feeds.findOne(url: feedUrl)
```
arnoldbm: A check to see if the feed exists on another protocol would be a nice enhancement here. Http vs Https.

jgoley:
@arnoldbm b322a9b takes care of your request and @nathanathan, it takes care of only throwing exceptions if on the server.

nathanathan:
👍

PR #174, opened 2016-11-18
Limit curator inbox date range to prevent overload

nathanathan:


arnoldbm:
Thoughts on updating the UI based on selected start date to disable end date past X days versus popping an alert?

Broham:
I see @arnoldbm's point regarding enforcing this through the UI rather than alerting them afterwards.  The code looks good so if we decide to go forward with the current method this PR looks good :+1:

PR #175, opened 2016-11-18
Style incident report details for small screens

jgoley:
Fixes styles of details in the incident report and locations tables.


nathanathan:
👍

PR #176, opened 2016-11-21
Show button to add incident reports when none exits

jgoley:
PT Bug: https://www.pivotaltracker.com/story/show/134619187


nathanathan:
The article url selector is not working in the incident report model opened by the summary addIncidentReport button.

jgoley:
@nathanathan My bad. Looks like I forgot to pass the articles and was showing the wrong incident report modal. 6a053ad fixes those issues and reverts the buttons back to primary — I thought, as they were, users might think they were disabled.

nathanathan:
👍

PR #178, opened 2016-11-21
Implement feed deletion

jgoley:
PT story: https://www.pivotaltracker.com/story/show/134515971

### DIFF
--- a/methods/feeds.coffee
+++ b/methods/feeds.coffee
@@ -21,3 +21,6 @@ Meteor.methods
     feed.addedByUserName = user.profile.name
     feed.addedDate = new Date()
     Feeds.insert feed
+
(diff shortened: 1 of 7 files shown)


methods/feeds.coffee
```diff
@@ -21,3 +21,6 @@ Meteor.methods
     feed.addedByUserName = user.profile.name
     feed.addedDate = new Date()
     Feeds.insert feed
+
+  removeFeed: (feedId) ->
+    Feeds.remove feedId
```
nathanathan: We should check that the user is an admin.

jgoley: Good call, 2ddb854 takes care of this.

nathanathan:
👍

PR #181, opened 2016-11-22
Style incident report modal

jgoley:
This improves the UX and styling of the incident report modal (to closer match the mockups) and reworks/refactors things a good bit.
- abandons the `suggestedIncidentModal` template and controller
- moves the `incidentForm` html and logic into the `incidentModal` template/controller. So there are 3 different states: adding, editing, and editing suggested.
- changes the dropdown selectors to "pill selectors"
- restructures the related files — add related files to a directory

PT chore: https://www.pivotaltracker.com/story/show/133696063


jgoley:
@nathanathan As we discussed, I reverted the change combining templates and logic in an earlier commit.
I had to create an object of reactive variables to pass down to the modal body template (`incidentForm`) [here](https://github.com/ecohealthalliance/eidr-connect/blob/d1ccb9aadf96f387183243e102e2b2b2c8b3f7d4/client/controllers/incidentReports/suggestedIndidentModal.coffee#L7-L11) and [here](https://github.com/ecohealthalliance/eidr-connect/blob/d1ccb9aadf96f387183243e102e2b2b2c8b3f7d4/client/controllers/incidentReports/incidentModal.coffee#L4-L8).
Is the name of that object appropriate? (`incidentFormDetails`)

nathanathan:
The cumulative checkbox does not appear to be saving/loading data.

I think the object name is good, but I don't like that we're passing two objects to the form to populate it and using two different methods to retrieve values from it. I think it would be best to use a single ReactiveVar/ReactiveDict to populate all the fields and retrieve the full form state from that. We can postpone that change though.

The new styles look great!

jgoley:
That's a good call with using a single data source. I suppose we could, in the future, use checkboxes or radios and style them as they are. That way all the data will remain in the form. I should have approached this by modifying the form styles. Sorry about that.
86024ff fixes the broken cumulative checkbox.

nathanathan:
👍

PR #185, opened 2016-12-02
Fix map tooltip link

jgoley:
Fixes [this bug](https://www.pivotaltracker.com/story/show/135367459) causing some links to fail in popups (when events have been selected on the left) and takes care of [this chore](https://www.pivotaltracker.com/story/show/134291919) (styling the popups)


nathanathan:
👍

PR #187, opened 2016-12-02
Allow admins to grant curator role to self

jgoley:
Fix list of curators

[PT BUG](https://www.pivotaltracker.com/story/show/135354941)


nathanathan:
👍

PR #188, opened 2016-12-05
Customly styled form elements

jgoley:
This reworks how we receive information from the Incident Report form. I recently changed things up to use reactiveVars to track the values of certain properties (when switching the select inputs to button-based checkboxes) Those buttons are now styled radio and checkbox inputs which allow us to receive all submitted data through the form.

### DIFF
--- a/.config/chimp.js
+++ b/.config/chimp.js
@@ -24,14 +24,14 @@ module.exports = {
   singleSnippetPerFile: true,
   recommendedFilenameSeparator: '_',
   chai: false,
-  screenshotsOnError: false,
+  screenshotsOnError: true,
   screenshotsPath: '.screenshots',
   captureAllStepScreenshots: false,
-  saveScreenshotsToDisk: true,
+  saveScreenshotsToDisk: false,
   // Note: With a large viewport size and captureAllStepScreenshots enabled,
   // you may run out of memory. Use browser.setViewportSize to make the
   // viewport size smaller.
-  saveScreenshotsToReport: false,
+  saveScreenshotsToReport: true,
   jsonOutput: null,
   coffee: true,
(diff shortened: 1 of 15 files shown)


jgoley:
I just noticed an issue where the other value is replaced with 'other' when editing. Fixing now.

arnoldbm:
I'm getting a routing error when trying to test this locally.
 
screen shot 2016-12-07 at 11 22 08 am

jgoley:
@nathanathan I tried adding tests but I think our current setup needs work before proceeding.
Since we are importing collections we probably should abandon the fixture package — I experienced some difficulty importing necessary files to create a fixture data.
I also tried to call methods directly within the test steps, but the user does not seem to be in the database...when executing the method.
Maybe we can use the `Meteor.isTest` flag like you've mentioned to import fixtures?
I've updated the tests in #191, fixing some issues I encountered.

jgoley:
@arnoldbm That's strange, I'll check it out.

jgoley:
@nathanathan Or, temporarily, i guess, we can test this in the browser? I was trying to create events and sources programmatically instead of in the browser.

jgoley:
@nathanathan I've updated the tests so they work with the changes to the form. I had trouble with locations populating while testing but maybe its something wrong with my environment.
@arnoldbm are you still experiencing the router issue?

jgoley:
There must be some sort of bug with phantomjs. The tests are passing when I choose chrome as the browser. For some reason, the incident report type is not being set while testing with phantomjs. There are a few issues about phantomjs not handling clicks on labels well (https://github.com/ariya/phantomjs/issues/12032) but the 'case' radio button has the checked attribute when the form is submitted so I don't know what's going on.

PR #190, opened 2016-12-06
Fix scatter plot update

dan-nyanko:
Fixes "EIDR-C incident plot doesn't update when incident is added"

This would happen when a user added an incident outside the plot's current _domain filter.  The fix  was to keep track of when new groups are added during the merge operation so that the domain filter can be temporarily disabled.

Integration tests have been added.

https://www.pivotaltracker.com/story/show/135206307

### DIFF
--- a/.meteor/packages
+++ b/.meteor/packages
@@ -2,7 +2,7 @@
 #
 # 'meteor add' and 'meteor remove' will edit this file for you,
 # but you can also edit it by hand.
-
+eidr:fixtures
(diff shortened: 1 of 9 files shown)


dan-nyanko:
commit [af6dee5] removed the method ScatterPlot.applyFilters as it would just call draw() without any params. Calling draw() without any params will redraw the plot and apply any added/removed filters to the existing data.

arnoldbm:
Looks good!

PR #191, opened 2016-12-07
Fix broken tests

jgoley:
I was experiencing some issues running `npm run chimp-test` and discovered, in the event there are issues with fibers, we need to prepend with `meteor` so run `meteor npm run chimp-test`

Stops restoring the database—seems like we should start with a clean slate after each test?
Adds shared or common steps.
Re-words things a little.
Moves checks for toasts and content from individual steps into shared steps / features
Adds tagging support to script
Adds chai
Kills phantomjs after tests run


dan-nyanko:
@jgoley I just noticed that this PR removes restoring the db after each test. By *not* restoring, each consecutive test will change the database and will eventually cause inconsistencies and tests will seemingly fail by random. 

Please explain "seems like we should start with a clean slate after each test" because thats exactly what the before and after decorators are doing [here](https://github.com/ecohealthalliance/eidr-connect/blob/master/tests/features/step_definitions/admin_steps.coffee#L7-L11).

jgoley:
@dan-nyanko My bad. For some reason the files in my dump are not empty. Must have been from a while ago. So I thought we were restoring data. I'll revert that change.

dan-nyanko:
Sorry I'm not explaining this correctly.  We are starting with a single dump with data as the 'clean' slate.  A clean slate meaning that all tests are run against the same exact snapshot of data.  This provides isolation and consistency for all tests and avoid any time consuming process of generating random mock data.

jgoley:
Ohh. I see. Sorry for the confusion. I suppose we may not need to worry with looking into `isTest` for now since we will be able to add sources and incident reports to the existing data.

nathanathan:
👍

PR #192, opened 2016-12-08
Fix events header UI on small screens

jgoley:
Fixes a small visual bug that was causing the elements in the events header (the "tracked events" heading, button, and search input) to break into incorrect positions.

[PT Bug](https://www.pivotaltracker.com/story/show/135662111)


nathanathan:
👍

PR #194, opened 2016-12-08
Add feature to display user status for curator inbox

dan-nyanko:
This PR adds the meteor package `mizzao:user-status` to keep track of which users are online.  It adds a property to the user document called 'status'.  The client then runs reactive queries against users whose status.online == true and curatorInboxSourceId == the current selectedSourceId.

The result of the queries is a count.  If the count is greater than 1 (accounting for their own status), then a bootstrap badge is displayed with the number of online users within the header of the `Associated Events`.

When the template isRemoved, the meteor method that sets curatorInboxSourceId is called again with null, to update that the user is no longer active on that curator source. 

The server also guards against users who loose connectivity and go offline by performing a server-side observe of status.online and setting curatorInboxSourceId = null for any user who goes offline.

image

To test:

Open two Chrome incognito browsers and login with separate accounts then navigate to curator inbox and experiment selecting sources with both browser windows open.

https://www.pivotaltracker.com/story/show/134787043

### DIFF
--- a/server/accounts.coffee
+++ b/server/accounts.coffee
@@ -2,17 +2,15 @@ Accounts.emailTemplates.siteName = "EIDR-Connect"
 Accounts.emailTemplates.from = "EIDR-C <no-reply@eha.io>"
 
 Meteor.startup ->
-  user = Meteor.users.findOne({'profile.name': 'Admin'})
-  if user
-    return
-  userData = Meteor.settings.private.initial_user
-  if userData
-    userData.profile = { name: 'Admin' }
-    console.log "[ Creating initial user with email #{userData.email} ]"
-    Accounts.createUser userData
-    newUserRecord = Meteor.users.findOne('emails.address': userData.email)
-    if newUserRecord
-      Roles.addUsersToRoles(newUserRecord._id, ['admin'])
-  else
-    console.warn '[ Meteor.settings.private.initial_user object \
-                        is required to create the initial user record ]'
+  unless Meteor.users.find().count()
+    userData = Meteor.settings.private.initial_user
+    if userData
+      userData.profile = { name: 'Admin' }
+      console.log "[ Creating initial user with email #{userData.email} ]"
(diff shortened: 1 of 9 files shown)


server/publications.coffee
```diff
@@ -42,3 +42,7 @@ Meteor.publish 'articles', (query={}) ->
 
 Meteor.publish 'feeds', ->
   Feeds.find()
+
+# User status
+Meteor.publish 'userStatus', () ->
+  Meteor.users.find({'status.online': true }, {fields: {'profile': 1, 'status': 1 }, transform: null})
```
dan-nyanko: I was thinking about doing the following with profile.name but did not.

image

The `transform: null` can increase the speed of the find operation by avoiding transformation layers by 3rd party plugins. In my experience, this is extremely noticeable when a ORM is used, which I'm not sure if we're using one or not in the project, so I say this is defensive programming on my part.

dan-nyanko:
Rebased w/ master

server/accounts.coffee
```diff
@@ -2,15 +2,17 @@ Accounts.emailTemplates.siteName = "EIDR-Connect"
...
-      userData.profile = { name: 'Admin' }
-      console.log "[ Creating initial user with email #{userData.email} ]"
-      Accounts.createUser userData
-      newUserRecord = Meteor.users.findOne('emails.address': userData.email)
-      if newUserRecord
-        Roles.addUsersToRoles(newUserRecord._id, ['admin'])
-    else
-      console.warn '[ Meteor.settings.private.initial_user object \
-                          is required to create the initial user record ]'
+  user = Meteor.users.findOne({'profile.name': 'Admin'})
```
nathanathan: What was the reason checking for an user named Admin instead of checking for no users? This could cause problems on the live instance since there is no user named Admin.

nathanathan:
👍 Looks good!

PR #201, opened 2016-12-13
adding soft delete logic

Broham:
Marking userEvent records as deleted instead of actually deleting them.


nathanathan:
The query in the server router also needs to be updated.

Broham:
@nathanathan - ok, added

nathanathan:
I noticed a couple other issues. The deleted events are still showing up in the suggested events list on the curator inbox. Also, the event pages for the deleted events are still accessible. I think they should only display a message like "This event has been deleted."

nathanathan:
👍 Looks good

PR #202, opened 2016-12-13
Clear searches

jgoley:
I went ahead and created a single template for all the search inputs. I figured it would reduce a fair amount of repetition with clearing inputs and force us into a consistent search UI/UX.

The template receives an object with options including:
- `textFilter`: either a `reactiveVar` or `ReactiveTableFilter`
- `id` the id associated with the element and `ReactiveTableFilter` if its passed
- `props` if a `ReactiveTableFilter` is not passed, one will be created with the `id` and `props` which is an array
- `toggleable`If set to true, the input will be initially hidden and appear when the search icon is clicked. When the user clicks on the search icon within the input, the input returns to its hidden state.
- `placeholder` defaults to 'Search'
- `classes` classes which will be applied to the input's parent element

### DIFF
--- a/client/controllers/maps/mapFilters.coffee
+++ b/client/controllers/maps/mapFilters.coffee
@@ -36,13 +36,17 @@ Template.mapFilters.onRendered ->
         varQuery._id = {$in: eventIds}
         filters.push(varQuery)
 
-    userSearchText = Template.instance().userSearchText.get()
-    nameQuery = []
-    searchWords = userSearchText.split(' ')
-    _.each searchWords, -> nameQuery.push {eventName: new RegExp(userSearchText, 'i')}
-    filters.push $or: nameQuery
+    userSearchText = instance.userSearchText.get().$regex
+    if userSearchText
+      nameQuery = []
+      searchWords = userSearchText.split(' ')
+      _.each searchWords, -> nameQuery.push {eventName: new RegExp(userSearchText, 'i')}
+      filters.push $or: nameQuery
 
-    Template.instance().data.query.set({ $and: filters })
+    if filters.length
+      instance.data.query.set({ $and: filters })
+    else
+      instance.data.query.set({})
 
 Template.mapFilters.helpers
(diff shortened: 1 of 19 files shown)


client/controllers/maps/mapFilters.coffee
```diff
@@ -75,19 +79,24 @@ Template.mapFilters.helpers
...
 Template.mapFilters.events
   'cancel.daterangepicker': (e, instance) ->
     $(e.target).val("")
     setVariables instance, 'on', []
 
-  'input .map-search': _.debounce (e, templateInstance) ->
-    e.preventDefault()
-    text = $(e.target).val()
-    templateInstance.userSearchText.set(text)
+  # 'input .map-search': _.debounce (e, templateInstance) ->
```
Broham: Should we just yank the commented out code below?

client/controllers/searchInput.coffee
```diff
@@ -0,0 +1,58 @@
...
+    classes:     Classes which will be applied to the input's parent element
+####
+
+clearSearch = (instance) ->
+  instance.textFilter.set('')
+  instance.$('.search').val('')
+
+Template.searchInput.onCreated ->
+  instanceData = @data
+  searching = true
```
Broham: could these lines be changed to: `searching = not instanceData.toggelable`?

jgoley: I think that would work if we always pass the toggelable property.
Maybe `searching = not instanceData.toggelable?` would work? But may be confusing.

client/controllers/maps/mapFilters.coffee
jgoley: yep, my bad

client/controllers/searchInput.coffee
Broham: hrrm - yeah I didn't think of that case.  I'd probably just leave it the way it is then.

Broham:
looks good :+1:

PR #203, opened 2016-12-13
Add disease field to event creation modal

nathanathan:


client/controllers/events/createEventModal.coffee
```diff
@@ -3,17 +3,20 @@
...
     unless valid
       toastr.error('Please specify a valid name')
       eventName.focus()
       return
-    newEvent = eventName.value
-    summary = target.eventSummary.value
-
-    Meteor.call 'editUserEvent', null, newEvent, summary, (error, result) ->
+    Meteor.call 'upsertUserEvent', {
+      eventName: eventName.value.trim()
```
Broham: Do we want to move this `trim` to line 8 for consistency sake?

imports/schemas/userEvent.coffee
```diff
@@ -4,22 +4,32 @@ userEventSchema = new SimpleSchema(
     optional: true
   articleCount:
     type: Number
+    optional: true
```
Broham: [I believe that](https://github.com/aldeed/meteor-simple-schema#optional) `optional: true` is the default in simple schema, are all of these required?

methods/userEvents.coffee
```diff
@@ -1,27 +1,26 @@
...
-            disease: disease
-            lastModifiedDate: now
-            lastModifiedByUserId: user._id
-            lastModifiedByUserName: user.profile.name
-          $setOnInsert:
-            creationDate: now
-            createdByUserId: user._id
-            createdByUserName: user.profile.name
-            articleCount: 0
+  upsertUserEvent: (userEvent) ->
```
Broham: I noticed that before we were trimming name, before we upserted.  Just wanted to confirm that we no longer need to do this.

imports/schemas/userEvent.coffee
nathanathan: The default is for them to be required. The keys are not required by the meteor method using this schema for validation.

methods/userEvents.coffee
nathanathan: I moved the trimming to the call side.

imports/schemas/userEvent.coffee
Broham: Oh, you are right, I totally misread that, sorry

methods/userEvents.coffee
Broham: ok sounds good

Broham:
Looks good :+1:

PR #206, opened 2016-12-13
add feature to confirm closing suggestedIncidentsModal

dan-nyanko:
https://www.pivotaltracker.com/story/show/134380381

To test:

1.  Login to the app and navigate to [Tracked Events]
2. Click on one of the events in the table
3. Click on the [Add Sources] button
4. Choose a `Suggested Article` by clicking on the text then click [Save]
5. Wait for the article to be processed...
6. Click on a red underlined text, wait for the next modal, then click [Confirm]
7. Note: that the number of accepted incidents has changed
8. Try to cancel/close the modal.
9. Note: a confirmation modal appears

### DIFF
--- a/client/controllers/cancelConfirmation/cancelConfirmationModal.coffee
+++ b/client/controllers/cancelConfirmation/cancelConfirmationModal.coffee
@@ -1,0 +1,3 @@
+Template.cancelConfirmationModal.helpers
+  data: ->
+    Template.instance().data
--- a/client/controllers/cancelConfirmation/cancelConfirmationModalBody.coffee
+++ b/client/controllers/cancelConfirmation/cancelConfirmationModalBody.coffee
@@ -1,4 +1,5 @@
 Template.cancelConfirmationModalBody.events
   'click .confirm': (event, instance) ->
+    instance.data.hasBeenWarned.set(true)
     instance.data.modalsToCancel.forEach (id) ->
       $("##{id}").modal('hide')
--- a/client/controllers/cancelConfirmation/cancelConfirmationModalBody.coffee
+++ b/client/controllers/cancelConfirmation/cancelConfirmationModalBody.coffee
@@ -1,0 +1,4 @@
+Template.cancelConfirmationModalBody.events
+  'click .confirm': (event, instance) ->
+    instance.data.modalsToCancel.forEach (id) ->
+      $("##{id}").modal('hide')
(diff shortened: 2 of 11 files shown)


Broham:
@dan-nyanko I was able to click outside of the modal to close it without displaying the modal warning.  Thinking we should probably display it for that use case as well?

Broham:
@dan-nyanko - are we sure we want to remove the functionality that allows people to dismiss modals by clicking off of them?  This is pretty standard functionality and I feel like we deliberately added this a while back.

jgoley:
I'd suggest we use a warning icon and color for this confirmation modal since they aren't really deleting the incidents. Maybe this can be an option passed to the confirmation template? Maybe two states: "warning" and "deletion". When a state of "warning" is passed in, we apply a warning class which changes the UI.
I'm sure we can implement this in another story/PR.

dan-nyanko:
@Broham In this instance, I'm sure that it should be disabled. Whats the purpose of allowing them to click the backdrop only to receive yet another modal that asks them to confirm if they want to abandon their changes.

dan-nyanko:
@jgoley These modals are similar but do have separate functionality. I'm not sure what we gain by combining the code into a single modal with a bunch of branching conditional logic.  E.g.: we'd need a conditional for the icon, a conditional for the action, etc. 

IMO it will be easier to maintain two separate modules that are similar but have separate functionality.  

Seem to be splitting a lot of hairs on this one.

Broham:
@dan-nyanko - my thought was that since the intent is the same whether you are clicking the "X", the close button or off the modal that the behavior should be the same for all three.  @jgoley - do you have any thoughts?

jgoley:
@dan-nyanko Keeping things separate sounds good. Can we go ahead and update the warning icon (maybe `fa-exclamation-triangle` with the warning color [here](https://github.com/ecohealthalliance/eidr-connect/blob/master/imports/stylesheets/variables.import.styl#L24)) and coloring of the button within the `cancelConfirmationModal` switching to `btn-warning` should do the job?

About the modal dismissal question: I do see the argument for forcing users to cancel within the modal but I don't see any harm in adding the _click the backdrop to dismiss_ behavior back to cancelConfimationModal since the behavior is the same as clicking cancel and users are generally trained into this pattern.

dan-nyanko:
@Broham @jgoley I'll add additional code that will allow the user to attempt dismissing the modal by clicking the backdrop and change the icon.

Broham:
Looks good :+1:

PR #208, opened 2016-12-14
fix curatorInbox date filter bug

dan-nyanko:
This happens when the user click on a startDate and does not choose an endDate, then clicks [Apply]. The code would see endDate is null and then set endDate as an object reference to startDate, thus startDate === endDate.  The fix is to make endDate as a copy of startDate and set its time to a millisecond before midnight.

- make the endDate a copy of startDate at a millisecond before midnight

https://www.pivotaltracker.com/story/show/134378881


nathanathan:
👍

PR #210, opened 2016-12-15
Fix unscrollable modals

jgoley:
The problem ([documented in this bug](https://www.pivotaltracker.com/story/show/136069771)) was happening because the `modal-open` class was being removed from the body when the _add source_ modal closed but after the _suggested incident report_ modal added it back.

Makes the _suggested incidents report_ modal stretch to the height of the viewport (and gives a min-height) and adds gradients on the top and bottom of the text window to imply additional content beyond.


nathanathan:
👍

PR #211, opened 2016-12-16
Inline form validation

jgoley:
~~*Need to fix all the broken tests~~
Some of the forms required a little finagling since we often listen for click events of buttons outside of the form instead the form's submission event. 

[PT Story](https://www.pivotaltracker.com/story/show/134791781)

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,18 +47,6 @@ meteor npm run-script start-test-server
 ./start-test-server.sh --app_port=3001 --mongo_host=127.0.0.1 --mongo_port=27017 --test_db=eidr-connect-test --prod_db=eidr-connect
 ```
 
-Tags
-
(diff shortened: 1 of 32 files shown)


nathanathan:
Saving edits on the event sources form isn't working.

nathanathan:
On the incident form the cumulative checkbox isn't being saved when checked. Also, when editing an incident, the location field will be shown as invalid unless it is modified.

jgoley:
@nathanathan I've fixed the bugs related to editing sources and incident reports.
There seems to be an issue with the validation when a location is not selected and the field is focused. The `has-error` class is not applied until the input looses focus. I'm not sure what is causing it but will look into it.

nathanathan:
👍

PR #213, opened 2016-12-19
Limit suggested article dates to before publication date

nathanathan:
Omit dates that are entirely in the future and truncate date ranges that extend into the future.


client/controllers/incidentReports/suggestedIncidentsModal.coffee
```diff
@@ -147,6 +138,18 @@ Template.suggestedIncidentsModal.onCreated ->
             Object.keys(timeAnnotation.timeRange.end).length +
             Object.keys(timeAnnotation.timeRange.end).length
           )
+          timeAnnotation.beginMoment = moment.utc(timeAnnotation.timeRange.begin)
+          # Round up the to day end
+          timeAnnotation.endMoment = moment.utc(_.extend({
```
arnoldbm: I think we can use the builtin .endOf('day') in momentJS here instead of rounding up like this

http://momentjs.com/docs/#/manipulating/end-of/

PR #214, opened 2016-12-19
Mark a source reviewed when leaving if has assocaiated events

jgoley:
Adds a reactiveVar which tracks if an event has been added. I so and the user selects another source, the source is marked reviewed and the next, selected, source is shown.

[PT Story](https://www.pivotaltracker.com/story/show/134378559)

### DIFF
--- a/client/controllers/curatorInbox/curatorEvents.coffee
+++ b/client/controllers/curatorInbox/curatorEvents.coffee
@@ -3,19 +3,22 @@ UserEvents = require '/imports/collections/userEvents.coffee'
 CuratorSources = require '/imports/collections/curatorSources.coffee'
 Articles = require '/imports/collections/articles.coffee'
 
+_getSourceId = (instance) ->
+  CuratorSources.findOne(instance.data.selectedSourceId.get())._sourceId
+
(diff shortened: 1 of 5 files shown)


nathanathan:
This breaks the undo review feature because there is no way to move off the article without causing it to be marked as reviewed. I recommend marking articles as reviewed at the time the user associates an event.

jgoley:
Oh, good point. I'll change it up.

jgoley:
@nathanathan I've updated this to mark the article as reviewed when an event is associated.
The UX is a little weird because the article is removed from the list on the left when an event is added. I suppose to prevent this we'd have to add some sort of temporary flag on the source when the user adds an event that marks it as reviewed only in the right-hand side, then when the user navigates away, mark it reviewed. Maybe too much trouble though.

Also, I was wondering if when a user removes all associated events, we should mark the source as un-reviewed?

nathanathan:
I think this is in a good state and will go ahead and merge. I don't think we should worry about automatically un-reviewing posts. I think this will mostly resolve the other issue: https://www.pivotaltracker.com/story/show/136512391

PR #215, opened 2016-12-20
fix confirm modal appearing after suggested incidents have been saved 

dan-nyanko:
Fixes https://www.pivotaltracker.com/story/show/136212851

The `incidentCollection` is now emptied after successful save to the remote.


nathanathan:
👍 Looks good

PR #216, opened 2016-12-20
Fix bug where associated suggested events not removed from table

jgoley:
Looks like at some point the instantiation of the `suggestedEventsHeaderState` reactiveVar made its way into an `@autorun` so it's value was being reset when a collection updated.

[PT Bug](https://www.pivotaltracker.com/story/show/136383865)


nathanathan:
👍 Looks good but there is a conflict

PR #217, opened 2016-12-20
fix title in tooltip of selected source

dan-nyanko:
https://www.pivotaltracker.com/story/show/136382371

- use a helper to dynamically update the tooltip title


nathanathan:
👍

PR #218, opened 2016-12-20
Update createInlineDateTangePicker to abbreviate months

jgoley:
Before this update, the abbreviation (down to 1 letter) was replaced after the user selected a new month. Modifies the click events of next and prev buttons to abbreviate the months

[PT Bug](https://www.pivotaltracker.com/story/show/136389589)


nathanathan:
👍

PR #219, opened 2016-12-22
Add tests for sources

jgoley:
This starts test for sources.
I have a few questions:
- Should we assume a connection to the SPA server or the associated db? [Here](https://github.com/ecohealthalliance/eidr-connect/blob/a75626f1defbc263159af3d1be5d1cedf91f434d/tests/features/step_definitions/sources_steps.coffee#L40) I kind of stub out the date and time since it depends on said connection. Is this the path we should take?
- Should I go ahead and add tests for reviewing incident reports from the newly added source?
- Do we need the chimp logs?

### DIFF
--- a/.config/chimp.js
+++ b/.config/chimp.js
@@ -31,7 +31,7 @@ module.exports = {
   // Note: With a large viewport size and captureAllStepScreenshots enabled,
   // you may run out of memory. Use browser.setViewportSize to make the
   // viewport size smaller.
-  saveScreenshotsToReport: true,
+  saveScreenshotsToReport: false,
   jsonOutput: null,
   coffee: true,
   compiler: 'coffee:coffee-script/register',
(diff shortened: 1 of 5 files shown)


nathanathan:
> Should we assume a connection to the SPA server or the associated db? Here I kind of stub out the date and time since it depends on said connection. Is this the path we should take?`

Manually filling in the data they way you do is a good approach. Eventually we could adding some SPA data to the test db dump and use that to test stuff that relies on SPA, but I don't think it's a priority.

> Should I go ahead and add tests for reviewing incident reports from the newly added source?

That sounds good. I recommend using a fixture to substitute the meteor method for getting grits enhancements with one that returns the same data every time.

> Do we need the chimp logs?

For the Jenkins tests I would leave in all the logging. If the tests work we can ignore it and if something goes wrong it can help us track down the issue.

nathanathan:
👍

PR #220, opened 2016-12-22
Fix source date input selector and trimming month

jgoley:
Fixes [this bug](https://www.pivotaltracker.com/story/show/136507507) and I figured out a better way of changing the day abbreviations.

### DIFF
--- a/client/controllers/sourceModal.coffee
+++ b/client/controllers/sourceModal.coffee
@@ -40,18 +40,17 @@ Template.sourceModal.onCreated ->
   @formValid = new ReactiveVar false
 
 Template.sourceModal.onRendered ->
+  publishDate = @timezoneFixedPublishDate
   pickerOptions =
-    format: 'M/D/YYYY'
-    inline: true
-    useCurrent: false
     singleDatePicker: true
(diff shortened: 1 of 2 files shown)


client/controllers/sourceModal.coffee
```diff
@@ -192,7 +191,11 @@ Template.sourceModal.events
     input.value = @url
     titleInput = instance.find('#title')
     titleInput.value = @subject
+    $(input).trigger('input').trigger('input')
```
nathanathan: Why does this need to be triggered twice?

jgoley: I was wondering the same. I think it was originally introduced here: https://github.com/ecohealthalliance/eidr-connect/commit/8756b6f5a23b419ab4303c58bd4fb371ae414393#diff-96f74db91476ffbdb7217d25f79ce73bR180

jgoley: @nathanathan I went ahead and removed the second trigger method call.
This also fixes [the bug related to the source time not being autofilled](https://www.pivotaltracker.com/story/show/136737137)

nathanathan:
👍

PR #222, opened 2016-12-27
filtering out associated events from suggested events

Broham:
Right now events that are already associated show under suggested events


nathanathan:
👍

PR #223, opened 2016-12-28
Change incident metadata icons to text titles

jgoley:


nathanathan:
The vertical alignment looks slightly off to me:
screen shot 2016-12-28 at 4 10 08 pm

jgoley:
@nathanathan Good catch. Looks like there was a conflict in line-heights.

nathanathan:
👍

PR #225, opened 2016-12-28
Fix flickering user indicator badge

jgoley:
Also rethinks the indicator, changes it from a count badge to an eye icon and relocates it to next to the source title.
I think explicitly showing the count is less important than indicating there are other curators actively viewing the same source.
Also ensure the tooltip shows when the template/view is destroyed and added back.

[PT Bug](https://www.pivotaltracker.com/story/show/136513111)

### DIFF
--- a/.meteor/versions
+++ b/.meteor/versions
@@ -37,7 +37,7 @@ eidr:fixtures@0.0.1
 ejson@1.0.12
 email@1.1.17_1
 fastclick@1.0.12
-fortawesome:fontawesome@4.6.3
+fortawesome:fontawesome@4.7.0
 fuatsengul:leaflet@1.0.1
 geojson-utils@1.0.9
 hot-code-push@1.0.4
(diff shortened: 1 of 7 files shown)


nathanathan:
I am still able to reproduce the flickering bug by opening two tabs for one user on different articles. Also, I think the red eye icon looks too sinister. How about we use one of the fa-user icons like this: http://fontawesome.io/icon/user-circle-o/

jgoley:
Hmm. I wasn't seeing the flickering while taking a look in two different windows. I'll take another look.
Good point about the eye-icon. I'll update to the user icon.

jgoley:
@nathanathan I think I discovered why the icon was flickering. Looks like [these lines](https://github.com/ecohealthalliance/eidr-connect/blob/master/methods/userStatus.coffee#L4-L6) are triggering an endless autorun loop. So I changed the method to update the user with `Meteor.userID()`.

nathanathan:
👍

PR #226, opened 2016-12-28
Add visual indication when fields are auto-filled

nathanathan:
This also fixes the status field bug here:
https://www.pivotaltracker.com/story/show/136737257

### DIFF
--- a/client/controllers/incidentReports/incidentForm.coffee
+++ b/client/controllers/incidentReports/incidentForm.coffee
@@ -24,6 +24,7 @@ Template.incidentForm.onCreated ->
 
   incident = @data.incident
   if incident
+    @suggestedFields = incident.suggestedFields
     @incidentData = _.extend(@incidentData, incident)
     if incident.dateRange
       @incidentData.dateRange = incident.dateRange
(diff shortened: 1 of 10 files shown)


client/controllers/incidentReports/incidentForm.coffee
```diff
@@ -24,6 +25,7 @@ Template.incidentForm.onCreated ->
 
   incident = @data.incident
   if incident
+    @suggestedFields = incident.suggestedFields
```
jgoley: Should we move this up to line 19 like we do [here](https://github.com/ecohealthalliance/eidr-connect/blob/dfb5101952d7b1033fa873607f9e6b51d768b488/client/controllers/incidentReports/suggestedIndidentModal.coffee#L7)?

imports/stylesheets/incidents/incidentForm.import.styl
```diff
@@ -61,3 +61,10 @@
         margin-bottom -1px
         > a
           color $primary
+
+.suggested
+  .control-label::after
+  label.cumulative::after
+  .tabs .active a::after
+    content " *"
+    color lighten($primary, 50%)
```
jgoley: I'm worried since most users are accustomed to asterisks indicating a field is required there may be some confusion with using them to indicate autofill. Maybe we can change the background color of the `form-group`?
Maybe like so:
screen shot 2017-01-03 at 11 38 36 am
And we could also add an icon like this: http://fontawesome.io/icon/magic/ with a tooltip explaining it has been autofilled?

nathanathan:
Thanks @jgoley

PR #229, opened 2017-01-03
adding route to display contents of revision.txt file

Broham:
This will eventually be used to sync automatic deploys on the BSVE environment


nathanathan:
👍

PR #232, opened 2017-01-04
add feedId to fetchPromedPosts

dan-nyanko:
Adds filter and field for 'feedId' to fetchPromedPosts so that only main ProMED feed posts appear in the inbox.

https://www.pivotaltracker.com/story/show/136512391


nathanathan:
👍

PR #234, opened 2017-01-04
fix: hide loading indicator when results <= 1

dan-nyanko:
https://www.pivotaltracker.com/story/show/137046527

### DIFF
--- a/client/controllers/searchInput.coffee
+++ b/client/controllers/searchInput.coffee
@@ -23,6 +23,7 @@ Template.searchInput.onCreated ->
   @searching = new ReactiveVar(searching)
   @textFilter = instanceData.textFilter or new ReactiveTable.Filter(instanceData.id, instanceData.props)
 
+
 Template.searchInput.helpers
   searchString: ->
     Template.instance().textFilter.get()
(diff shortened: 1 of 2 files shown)


client/controllers/searchInput.coffee
```diff
@@ -44,6 +45,15 @@ Template.searchInput.events
...
         $options: 'i'
+      if instance.data.tableId
+        Meteor.setTimeout(->
+          count = parseInt($("##{instance.data.tableId}").next().find('span.rows-per-page-count').text(), 10)
+          if count == 0
+            # hide loading indicator
+            $('.loading').hide()
+          else
+            $('.loading').show()
+        , 0)
```
jgoley: Maybe we could use `Meteor.defer` instead of `Meteor.setTimeout`? I suppose it would be a little cleaner.

dan-nyanko: The loading indicator would be partially shown if there was one row in the table, so I changed it to be <= 1.  AFAIK meteor.defer is the same thing as `Meteor.setTimeout ... , 0` and the syntax is cleaner especially with coffeescript.

PR #235, opened 2017-01-05
rolling back revision changes

Broham:
These are no longer needed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,5 +16,3 @@ tests/dump/eidr-connect
 tests/log/
 
 *.pid
-
-**/revision.txt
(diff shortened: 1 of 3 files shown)


nathanathan:
I don't think there is a reason to get rid of this. It could be useful for debugging.

Broham:
@nathanathan - what do you think we would use it for for debugging?

nathanathan:
We could use it to verify the correct version of the app is deployed.

Broham:
Ok - lets leave it then.  I'll make sure the `requirements.txt` file is being created correctly.

PR #236, opened 2017-01-05
Curator inbox - link source to promed

jgoley:
Links the text of the source title to the promed source (opens in new tab).
This is promed centric currently since I've hard-coded the promed url.
Also sets up the title so that a tooltip is only shown if the title is truncated with an ellipsis.

PT Story: https://www.pivotaltracker.com/story/show/134287975


nathanathan:
👍

PR #242, opened 2017-01-09
Collapse long event summaries

nathanathan:


imports/stylesheets/mixins.import.styl
```diff
@@ -80,3 +80,16 @@ gradientTopAndBottom()
...
+  &::after
+    content ''
+    height 20px
+    position absolute
+    left 0
+    right 0
+    z-index 10
+  &::after
+    background linear-gradient(to bottom, alpha(white, 0%), white)
+    bottom 0
```
jgoley: There's a class called `gradient-bottom` [here](https://github.com/ecohealthalliance/eidr-connect/blob/343dc99aacea92a7f0761ccf76d523a5983e8891/imports/stylesheets/utils.import.styl#L35-L46). Maybe we could now call the new mixin in that class and pass in the height as a variable (with a default) for flexibility.

Also, we can remove that second `&::after` selector

jgoley:
@nathanathan I've made some updates you might want to review: sizing the summary container based on the height of the metadata container.

PR #247, opened 2017-01-10
Restrict editing/deleting of sources and IRs to admins

jgoley:
[PT Bug](https://www.pivotaltracker.com/story/show/137362759)


nathanathan:
👍

PR #252, opened 2017-01-11
Allow users to deselect incident status

jgoley:
[PT Story](https://www.pivotaltracker.com/story/show/137420161)


nathanathan:
👍

PR #253, opened 2017-01-11
Add 'approximate' property to IRs and display in details

jgoley:
I grouped the IR type inputs and the child inputs appear when a type is selected. I added a grouping of checkboxes called "Integrity" which includes `cumulative` and `approximate` (`cumulative` was included in "Additional information"). Not sure if its an appropriate title, let me know what you think.
screen shot 2017-01-11 at 1 41 27 pm

If `cumulative` or `approximate` are selected, they appear in the IR details UI like so:
screen shot 2017-01-11 at 1 37 23 pm

### DIFF
--- a/client/controllers/incidentReports/incidentForm.coffee
+++ b/client/controllers/incidentReports/incidentForm.coffee
@@ -105,6 +105,9 @@ Template.incidentForm.helpers
     if fieldName in Template.instance().suggestedFields.get()
       "suggested"
 
+  typeIsSelected: ->
+    Template.instance().incidentType.get()
+
 Template.incidentForm.events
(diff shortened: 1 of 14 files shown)


nathanathan:
The styling on the autofill background looks off here:
screen shot 2017-01-12 at 7 47 20 pm

For the group title I would suggest "Count Attributes" instead. Also, the group could be hidden when the incident type is "Other".

jgoley:
@nathanathan Good catch. 45ce5b2 updates the autofill styling:
- highlights the entire incident type block instead of individual inputs
- Fixes positioning issues with the autofill magic icon in the tabbed date-selector form-group

nathanathan:
👍

PR #254, opened 2017-01-11
append template instance data.index to create unique ID

dan-nyanko:
https://www.pivotaltracker.com/story/show/137299227


nathanathan:
👍

PR #256, opened 2017-01-12
Make suggested incident report dates use article timezone

nathanathan:
Fixes this bug: https://www.pivotaltracker.com/story/show/137420455


jgoley:
Looks good.
I found an unrelated bug which seems to indicate we need [that redundant method call](https://github.com/ecohealthalliance/eidr-connect/blob/76f41f7ccb894bda9fcaab14245a0266260baac0/client/controllers/sourceModal.coffee#L197) removed a while back. I registered a bug for it: https://www.pivotaltracker.com/story/show/137602203

PR #257, opened 2017-01-13
Fix updating datetime when clicking suggested source

jgoley:
For some reason setting the promed regex as a variable caused the `exec` function to alternately return null even when everything matched (Any ideas why this would be happening? The same behavior occurs in the console).
I also refactored the click event of suggested articles - removed the jquery code and update values of the fields using helpers.

### DIFF
--- a/client/controllers/sourceModal.coffee
+++ b/client/controllers/sourceModal.coffee
@@ -80,7 +80,8 @@ Template.sourceModal.helpers
       'save-modal'
 
   title: ->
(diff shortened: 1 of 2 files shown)


nathanathan:
When editing existing sources the titles are blank.

jgoley:
My bad, forgot the title is the `title` property after it is saved instead of `source`. a16fa87 fixes it.

nathanathan:
👍

PR #261, opened 2017-01-17
Add tooltips to explain incident report form fields

nathanathan:


jgoley:
I think we need to give the user some indication that additional info is available. I'll make this change.

jgoley:
I updated the incidentForm to show info icons indicating a description is available.
👍

PR #262, opened 2017-01-17
Source title autofill

jgoley:
This autofills the source title when source URLs are manually entered.
I branched off of #257 because it contains a few pertinent fixes.

It was registered as a [bug](https://www.pivotaltracker.com/story/show/137420353) but I'm not sure we implemented this in the past.


nathanathan:
👍

PR #263, opened 2017-01-17
Update Suggested IR UI - content on left and right

jgoley:
Breaks the UI out to a two-pane UI at 992px and above. Snippet on left and form on right:
screen shot 2017-01-17 at 4 15 17 pm

[PT story](https://www.pivotaltracker.com/story/show/137783155)


nathanathan:
👍 Looks good!

PR #264, opened 2017-01-17
Do not show source URL in IR review/confirmation

jgoley:
[PT Story](https://www.pivotaltracker.com/story/show/137783437)


nathanathan:
👍

PR #266, opened 2017-01-18
Fix ir location autofill bug

jgoley:
I noticed in some circumstances when a location was not determined by the GRITS api, the location field was still marked as autofilled and the user could submit the form.
This PR:
- If there is no location, removes the `location` field from the `suggestedFields` array and runs the validator to mark the field as invalid.
- Refactors the removal of suggestedFields into a function.
- Validates the suggested incident form

### DIFF
--- a/client/controllers/incidentReports/incidentForm.coffee
+++ b/client/controllers/incidentReports/incidentForm.coffee
@@ -160,7 +160,6 @@ Template.incidentForm.events
     _removeSuggestedProperties(instance, ['cumulative'])
 
   'submit form': (event, instance) ->
-    instance.$(event.currentTarget).validator()
     prevented = event.isDefaultPrevented()
     instance.data.valid.set(not prevented)
     if prevented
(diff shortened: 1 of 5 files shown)


client/controllers/incidentReports/incidentForm.coffee
```diff
@@ -70,10 +74,14 @@ Template.incidentForm.onRendered ->
...
-  @autorun ->
-    instance.incidentType.get()
-    Meteor.defer ->
-      instance.$('#add-incident').validator('update')
+  @autorun =>
+    @incidentType.get()
+    Meteor.defer =>
+      @$('#add-incident').validator('update')
+
+  if not @incidentData.locations.length
```
nathanathan: It would be better to prevent locations from being added to suggestedFields initially when there are no locations.

jgoley: Good call. How's [this for a solution](https://github.com/ecohealthalliance/eidr-connect/blob/ba0b45193343a2a8ef6a4ee0581863a525fe9f37/client/controllers/incidentReports/suggestedIncidentsModal.coffee#L219)?

nathanathan: Looks good

nathanathan:
:+1:

PR #267, opened 2017-01-19
Add extract incidents page

nathanathan:
The /extract-incidents page allows users to submit free-text articles and download a CSV of all the incidents extracted from them.


jgoley:
Should we allow users to close the "suggested incident report" modal without showing the "cancel confirmation" modal? Maybe only show the cancel confirmation when the user makes a change to one of the suggested IRs?

PR #270, opened 2017-01-19
Accessibility keyboard focus

jgoley:
This increases the keyboard accessibility for several areas in the application.
(Makes reactiveTable rows keyboard focusable by default and updates events to react to `enter` and `space`)


nathanathan:
👍

PR #272, opened 2017-01-23
Style autofilled source form inputs

jgoley:
This applies a variant of existing autofill styles to the source form. I thought that since the autofill occurs after a user action (clicking a suggested article) we can reduce the visual indication—just change the border and background of the inputs:
screen shot 2017-01-23 at 11 01 33 am


nathanathan:
👍

PR #276, opened 2017-02-01
fix hidden chart labels; update yLabel text

dan-nyanko:
https://www.pivotaltracker.com/story/show/138618367

image


nathanathan:
👍

PR #279, opened 2017-02-13
Improve UX of date filtration in curator inbox

jgoley:
- Fixes [this bug](https://www.pivotaltracker.com/story/show/139570855) causing the UI to hang on loading when the current date is selected for filtration by setting the max date to the publish date of the latest source in the collection
- Shows the default filtration range
- Only shows the reset button when a user has changed the range
- Refactors how we show content when range changes: no longer reload the calendar
- When user selects only one date, the UI updates to reflect single-date selection (lingering range highlighting is removed)
- Shows loading indicator over only the source list — related to [this story](https://www.pivotaltracker.com/story/show/139570825)


nathanathan:
👍

PR #284, opened 2017-02-21
Add disease suggestions to incident reports

nathanathan:


imports/stylesheets/variables.import.styl
```diff
@@ -79,6 +79,7 @@ $incident-type-colors = {
   'date': lighten($positive, 50%)
   'case': lighten($highlight, 60%)
   'uncertain': darken($l-gray, 5%)
+  'disease': #f1d9eb
```
jgoley: Looks like instead of explicitly setting the disease property to the hex value and because its the same as the case color (`lighten($highlight, 60%)`), we could use a variable. Maybe `$highlight-light`?

client/controllers/incidentReports/suggestedIncidentsModal.coffee
```diff
@@ -267,9 +272,15 @@ Template.suggestedIncidentsModal.onCreated ->
         if suspectedAttributes.length > 0
           incident.status = 'suspected'
         incident.url = [@data.article.url]
+        # The disease field is set to the last disease mentioned,
+        # document classification, or event disease with overides in that order.
         event = UserEvents.findOne(@data.article?.userEventId)
         if event?.disease
           incident.disease = event.disease
+        result.diseases.forEach ({name})->
```
nathanathan: That would create the disease field when there is no disease which would cause a problem in the suggested field highlighting. Speed isn't a problem here since all of those arrays have only a few items.

Broham: ahh ok - yeah :+1: then

imports/stylesheets/variables.import.styl
nathanathan: The disease color is actually a little different, it's more purple than red. Since the primary colors are taken, it's hard to make it more distinctive.

jgoley: Oh, I see now. Sorry about that.

PR #285, opened 2017-02-21
Update the position of the y-axis label

jgoley:
I suppose this probably won't be a problem because the bug only occurs when the counts are above 100,000 but I nudged the label to the left a little. If it does become a problem I suppose we could use a solution like this: https://github.com/d3/d3-axis/blob/master/README.md#axis_tickFormat


nathanathan:
👍

PR #286, opened 2017-02-22
Groundwork for Smart Events

nathanathan:
https://www.pivotaltracker.com/story/show/139701083

This adds basic pages, methods, publications, and a collection and schema for smart events. Most of the code is adapted from the user events code.

@jgoley Could you review this?

### DIFF
--- a/client/router.coffee
+++ b/client/router.coffee
@@ -83,6 +83,10 @@ Router.route "/user-events",
   name: 'user-events'
   title: 'User Events'
 
+Router.route "/smart-events",
+  name: 'smart-events'
+  title: 'Smart Events'
+
 Router.route "/curator-inbox",
   name: 'curator-inbox'
   title: 'Curator Inbox'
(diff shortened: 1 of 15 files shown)


client/router.coffee
```diff
@@ -106,6 +110,10 @@ Router.route "/user-event/:_id/:_view?",
     articles: Articles.find({'userEventId': @params._id}, {sort: {publishDate: -1}}).fetch()
     incidents: Incidents.find({'userEventId': @params._id}, {sort: {date: -1}}).fetch()
 
+Router.route "/smart-event/:_id/:_view?",
```
jgoley: I wonder if we will need `_view`?

jgoley:
I noticed the modal backdrop was sticking around on smart event creation. I made some updates in 79f5b33 which fix this issue.
Looks good otherwise 👍

client/router.coffee
nathanathan: If we're going to use tabs on the smart event page I think we should keep it, otherwise let's take it out.

PR #287, opened 2017-02-22
Smart events table

jgoley:
This branches of the yet-to-be-merged scaffolding branch (#286) . Sets of the table and related filtering mentioned in [this PT story](https://www.pivotaltracker.com/story/show/139700905).

### DIFF
--- a/client/controllers/curatorInbox/curatorInbox.coffee
+++ b/client/controllers/curatorInbox/curatorInbox.coffee
@@ -30,6 +30,15 @@ createNewCalendar = (latestSourceDate, range) ->
       # Update selection to indicate one date selected
       $(@).find('.start-date').addClass('end-date')
 
+uniteReactiveTableFilters = (filters) ->
+  reactiveFilters = []
+  _.each filters, (filter) ->
+    _filter = filter.get()
+    if _filter
+      reactiveFilters.push _.object filter.fields.map (field)->
+        [field, _filter]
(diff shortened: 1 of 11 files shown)


nathanathan:
The next page button isn't working for me. Looks good otherwise.

jgoley:
@nathanathan I'm not sure what's causing the bug with pagination. It's defintily related to the filter which shows only the events created by the current user. For some reason the currentPage is being increased then set back to zero. [This `autorun`](https://github.com/ecohealthalliance/eidr-connect/blob/7bb26a0ba3ef636af0f5c5297dc6e54317434ba2/client/controllers/smartEvents/smartEvents.coffee#L50-L55) is being called when the pagination buttons are clicked which is odd since I don't think the reactive variable within is updated when those buttons are clicked.
I'll continue to look into it but if you see anything please let me know.

nathanathan:
@jgoley That autorun might be retriggered by the onRender function being called rather than the variables changing. I suspect the pagination problem is related to this line: https://github.com/ecohealthalliance/eidr-connect/pull/287/files#diff-ee55dfcd4ecb8d9100c93124978d9aa7R18

jgoley:
@nathanathan I was over-engineering things by updating the filter in an `autorun`.  I moved that logic to the event so now it only happens when the _show my events_ button is clicked.

nathanathan:
👍 Looks good

PR #291, opened 2017-02-28
Smart events incident reports daterange

jgoley:
[PT story](https://www.pivotaltracker.com/story/show/139701469)
Should we allow the user to optionally add a daterange? Currently if they pass over the calendar, the range is set to the current date.


nathanathan:
@jgoley Are there any changes in this branch that we're included in the smart-events-incident-reports-location branch?

jgoley:
I branched #292 from this and didn't rebase after so everything in this should now be in master.

PR #292, opened 2017-02-28
Smart events incident reports locations

jgoley:
This builds on #291 and extends the smart events IR query to include IRs that have a location matching one of the event's locations.
Currently all the alternative names are searched. Seems excessive but several locations of IRs seem to be within the alt list. 

[PT Story](https://www.pivotaltracker.com/story/show/139701521)


nathanathan:
I don't think the alternate name matching is needed. It will match locations different from the one selected that happen to share a name. In order to include locations contained within the selected geoname, we should query for geonames with matching countryName, admin1Name and admin2Name fields. The featureCode can be used to determine if the selected geoname is a country or administrative division and add the appropriate fields to the query.

Also, I think we should move the summary field to the bottom of the smart event form and make it so none of the query fields are required.

jgoley:
Thanks @nathanathan I'll make these changes ASAP.

client/controllers/smartEvents/smartEvent.coffee
```diff
@@ -5,15 +5,39 @@ Incidents = require '/imports/collections/incidentReports'
...
+      if locations
+        locationNames = []
+        for location in locations
+          featureCode = location.featureCode
+          locationNames.push(location.countryName)
+          if featureCode is 'ADM1' or featureCode is 'ADM2'
+            locationNames.push(location.admin1Name)
+          if featureCode is 'ADM2'
+            locationNames.push(location.admin2Name)
+        query['locations.name'] = $in: _.uniq(locationNames)
```
nathanathan: This should match on the corresponding geoname properties rather than the location name. This is going to match the incident reports that mention a location containing the selected location rather than those contained by the selected location.

jgoley: It sure is. My bad. I overlooked making that change when updating the other parts.

nathanathan:
👍 Looks good

PR #293, opened 2017-03-01
Delete IRs of deleted events

jgoley:
This soft deletes IRs associated with deleted user events and deletes them when a user event is deleted to prevent these IRs showing with smart events.


nathanathan:
👍

PR #294, opened 2017-03-02
Update axes when plot data is updated

jgoley:
Fix for [this bug](https://www.pivotaltracker.com/story/show/140577847).
The yMax of the y-axis was not being updated when the data changed (e.g. when the cumulative button was toggled). Now when the plot us updated, the axes are as well.

### DIFF
--- a/imports/charts/ScatterPlot.coffee
+++ b/imports/charts/ScatterPlot.coffee
@@ -71,6 +71,9 @@ class ScatterPlot extends Plot
   ###
   draw: (data) ->
     super(data)
+    # if plot is not zoomed update the axes
+    unless @isZoomed()
+      @axes.update(@getGroupsNodes())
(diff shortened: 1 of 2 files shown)


nathanathan:
This change breaks the zoom feature for me.

jgoley:
This should be good to go now. I've updated the `update` method of the `ScatterPlot` class to only update the axes when a zoom area is not established.
I added a few methods to help with this:
`resetZoomArea`: Sets all properties of the `zoomArea` obj to 0 (method of `Zoom` class)
`isZoomed`: used to determine if the plot is in a zoomed (checks if all the values of `zoomArea` are 0 (method of `ScatterPlot` class)

Broham:
Looks like the zoom is working now - going to go ahead and merge

PR #295, opened 2017-03-06
Fix bug with showing IR edit form - get collection as array

jgoley:
Fixes [this bug](https://www.pivotaltracker.com/story/show/140877257).
Caused by a recent change where we pass a collection of articles to the form instead of an array.


Broham:
looks good :+1:

PR #296, opened 2017-03-07
showing source text by default

Broham:
Basically just removing conditional and toggle logic for the source text section


arnoldbm:
This looks good

jgoley:
👍

PR #297, opened 2017-03-08
Curator inbox get article enhancements

jgoley:
[PT Story](https://www.pivotaltracker.com/story/show/141164735)

Looks like there is currently no automated process for getting articles from ProMed—they are fetched when the inbox is loaded and added to the `CuratorSources` collection. So to move this along quickly, I've updated the inbox to process articles when they are clicked. The resulting incident reports are added to the collection (I made some fields in the schema optional since now some IRs may not have certain properties). The resulting UX is not really ideal because, I think, the processing time is a little too long for the user to wait. So we should think about automating this in the future.
I moved the logic associated with creating IRs from enhancements to a Meteor method so we can call it from multiple areas.
**This does not make the updates to the IR schema which track status.**

### DIFF
--- a/client/controllers/curatorInbox/curatorInbox.coffee
+++ b/client/controllers/curatorInbox/curatorInbox.coffee
@@ -100,7 +100,6 @@ Template.curatorInbox.onRendered ->
         console.log(err)
         return toastr.error(err.reason)
 
-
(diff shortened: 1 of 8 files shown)


Broham:
looks good :+1:

PR #298, opened 2017-03-08
Ir table

Broham:
Moving the Incident Report table to it's own template.  Reference the new template from the suggest incidents modal _and_ the curator inbox now

### DIFF
--- a/client/controllers/curatorInbox/curatorSourceDetails.coffee
+++ b/client/controllers/curatorInbox/curatorSourceDetails.coffee
@@ -29,6 +29,7 @@ _getSource = (instance, sourceId) ->
   instance.source.set source
 
 Template.curatorSourceDetails.onCreated ->
+  @incidentCollection = new Meteor.Collection(null)
   @notifying = new ReactiveVar(false)
   @source = new ReactiveVar(null)
   @reviewed = new ReactiveVar(false)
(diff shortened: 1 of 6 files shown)


client/controllers/curatorInbox/curatorSourceDetails.coffee
```diff
@@ -29,6 +29,7 @@ _getSource = (instance, sourceId) ->
   instance.source.set source
 
 Template.curatorSourceDetails.onCreated ->
+  @incidentCollection = new Meteor.Collection(null)
```
jgoley: This may not be necessary since its declared later in the autorun?

Broham: Yep - good call

jgoley:
Looks like on the extract-incidents route the table appears when the source is processing.
Also looks like you will need to pass in at least the articles? There is an error when clicking the case count which should open the modal to edit the IR. You may need to transfer some events and controller logic to the new template?

jgoley:
`Uncaught TypeError: this.data.articles.fetch is not a function`
Or maybe the articles is an array instead of a collection?

Broham:
@jgoley - good catch.  I ended up passing the flag that handles the showing/hiding of the IR table to the template itself.  Look better?

Broham:
Ok error should be resolved @jgoley - want to give it one more review before I merge?

Broham:
Ahh ok @jgoley - where else are you seeing it?

jgoley:
Lines 51 and 85 in the incidentForm controller.

Broham:
ok @jgoley - i pulled those out as well

jgoley:
👍  Looks good

PR #299, opened 2017-03-08
Change ir url to string

jgoley:
Updates the schema of Incident Reports—changes the `url` property from an array to a string.
Updates existing IRs in database and areas were the url is referenced, updated or presented.
[PT Chore](https://www.pivotaltracker.com/story/show/141256773)


Broham:
looks good :+1:

PR #301, opened 2017-03-08
Do not update selected source if currently selected

jgoley:
Small update to fix a bug associated with the recent change to curator inbox.
[PT Bug](https://www.pivotaltracker.com/story/show/141355003)


Broham:
Looks good 👍

PR #302, opened 2017-03-09
Curator inbox highlight ir related text

jgoley:
Adds the annotated source content and allows users to scroll to text related to IR by click table row.
Removes curator events from the inbox UI. I left the templates and controllers in case we want to use them in the future. Good idea?

[PT Story](https://www.pivotaltracker.com/story/show/141165917)


Broham:
@jgoley - when the inbox first loads I'm not seeing the source text for the article.  Is it showing for you?

jgoley:
@Broham Is the text not appearing always on load or on load and after the source is processed? It's loading okay for me.
Just noticed that the _uncertain_ IRs are not highlighted. I'll update to highlight those. (I think eventually those will be filtered from the table)
The saturation of the accepted IRs is increased when they are clicked and scrolled. Its probably too subtle but I didn't want to change the color too much.

screen shot 2017-03-10 at 5 58 43 pm

I wonder if we should go ahead and take care of the updates mentioned in [this chore](https://www.pivotaltracker.com/story/show/141456647)? If the right pane is scrolled, then a) the snippet won't be visible and b) I think the scroll amount will be incorrect because its based on the offset from the top of the window.

nathanathan:
ProMED is using the curator events to create timeline enhancements. So we don't have to start maintaining forks, I suggest preserving the old curator inbox page and adding a flag so we can display the old version on the instance they are using.

It seems like it would work better to make the incident report modal open when the incident table rows are selected rather than scrolling the article above since the modal allows the incident data to be edited and indicates the locations in the article that the meta-data comes from.

jgoley:
@nathanathan The task I'm currently working on is to allow curators to edit IRs. How about we change the UX so that by clicking the row, the edit modal appears (as you've proposed) and clicking some icon in the row, the source pane scrolls to the related text?

nathanathan:
That sounds good

jgoley:
I think this is good to go now.
Changes:
- Source is now scrolled on hover (after hovering for 500ms) instead of on click 
- I think fixes the issue with source content not showing. I'm having trouble reproducing but I think it was due to some IRs not having textOffsets (1793f59)
- Updates the UI so only the table scrolls
- Updates annotation highlight color so it's obviously different from others
- Adds media queries to show the text and table side-by-side at and above 1500px:
screen shot 2017-03-13 at 4 56 54 pm

nathanathan:
👍 Looks good

PR #304, opened 2017-03-10
adding ability to multiselect accept/reject incident reports

Broham:


client/controllers/incidentReports/incidentTable.coffee
```diff
@@ -0,0 +1,35 @@
...
+selectedIncidents = (event, instance) ->
+  # return all the selected incident-ids
+  $('table.incident-table tr.active td.count').map(() ->
+    $(this).data("incident-id")
+  ).get()
+
+changeIncidentStatus = (status) ->
+  incidentIds = selectedIncidents()
+  $.each incidentIds, (index, incidentId) ->
+    Meteor.call 'getIncidentReport', incidentId, (error, incident) ->
```
jgoley: I wonder if this method call is necessary. Could we update the IR and only set the accepted property?

jgoley:
I think we should go ahead and filter out the IRs that are not machine-accepted—consider and present IRs within the table as accepted. Then the main, and potentially only, action of the user within this area of the UI will be to reject IRs that were incorrectly marked as accepted? (And, down the road, to add them to events) When rejected they will be presented in the table of rejects (Salon des Refusés) mentioned in this story: https://www.pivotaltracker.com/story/show/141167197

client/controllers/incidentReports/incidentTable.coffee
Broham: @jgoley - I was thinking it wouldn't be needed initially too, but when I tried to call `editIncidentReport` it was throwing validation errors because nothing else for the IR was set.  I can play around some more to see if there is a way to set only the field I care about without specifying all the other fields.

Broham:
@jgoley - that seems like a big shift in the planned functionality.  Do you think we should put it out in front of the rest of the team at the meeting this morning and see what they think?

client/controllers/incidentReports/incidentTable.coffee
jgoley: I think that's because the IRs in the local collection immediately after processing have all the raw properties. Maybe we should change up the `createIncidentReportsFromEnhancements` to always return an object with the necessary properties picked? (https://github.com/ecohealthalliance/eidr-connect/blob/master/methods/incidentReports.coffee#L270-L273)
This may be the best way—increases consistency and decreases complexity (assuming we don't need the raw props somewhere else)

Broham:
@jgoley - ok, I moved the state out of the DOM and into a collection.  I also streamlined the update functionality so we don't have to retrieve the documents before updating them.  Going to take a look at the new "reject" functionality that we talked about next

Broham:
@jgoley - removed 'confirm' functionality as well.  If you have a moment take a look and let me know if you think everything looks good

client/controllers/incidentReports/incidentTable.coffee
```diff
@@ -0,0 +1,33 @@
...
+    Meteor.call 'updateIncidentReport', incident, false, (error, result) ->
+      if error
+        console.log "ERROR calling editIncidentReport", error
+        toastr.error("There was a problem updating your incident reports!")
+        return
+      toastr.success("Incidents reports udpated!")
+
+Template.incidentTable.helpers
+  showMultiActions: ->
+    Template.instance().rowsSelected.get()
```
jgoley: I think we can avoid tracking this with its own reactiveVar and within this helper check if `selectedIncidents` has a length.

client/controllers/incidentReports/incidentTable.coffee
```diff
@@ -0,0 +1,33 @@
+Template.incidentTable.onCreated ->
+  @rowsSelected = new ReactiveVar(false)
+  @selectedIncidents = new ReactiveVar([])
+
+changeIncidentStatus = (status, instance) ->
+  $.each instance.selectedIncidents.get(), (index, incidentId) ->
```
jgoley: In other areas of the app we've used either `forEach`, `_.each`, or `for...in`. Maybe we should switch to one of those so we can stay semi-consistent?

methods/incidentReports.coffee
```diff
@@ -68,6 +68,9 @@ parseSents = (text)->
 
 
 Meteor.methods
+  getIncidentReport: (incidentId) ->
+    Incidents.findOne({_id: incidentId})
```
jgoley: I think we can remove this now.

client/controllers/incidentReports/incidentTable.coffee
```diff
@@ -0,0 +1,33 @@
...
+        toastr.error("There was a problem updating your incident reports!")
+        return
+      toastr.success("Incidents reports udpated!")
+
+Template.incidentTable.helpers
+  showMultiActions: ->
+    Template.instance().rowsSelected.get()
+
+Template.incidentTable.events
+  'click table.incident-table tr': (event, instance) ->
```
jgoley: Lets make this more specific so clicks to the `thead` row won't trigger the event.

Broham:
@nathanathan - when you have a moment could you look at merging this?

nathanathan:
👍

PR #305, opened 2017-03-14
Implement IR editing in curator inbox

jgoley:
[PT Story](https://www.pivotaltracker.com/story/show/141164797)

This is currently not reactive since we are using a local collection created when the source is processed (because I wanted to ease into adding all IRs into the db — IRs are currently not added to the db in the event page flow until they are confirmed.) Maybe we update the curator inbox IR table to use the incidents collection while implementing a future story or add a chore to do so.

### DIFF
--- a/client/views/incidentReports/incidentTable.jade
+++ b/client/views/incidentReports/incidentTable.jade
@@ -8,7 +8,7 @@ template(name="incidentTable")
         | Deslect All
       else
         | Select All
-  table.table.incident-table(
+  table.table.featured.incident-table(
     hidden=annotatedContentVisible
     data-tableexport-display="always")
     thead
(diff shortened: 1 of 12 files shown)


client/views/incidentReports/incidentTable.jade
```diff
@@ -9,26 +9,20 @@ template(name="incidentTable")
         th Start Date
         th End Date
         th.locations-header Locations
-        th Status
```
nathanathan: Is it okay to remove these columns from the /extract-incidents page? Perhaps we should create a separate incident table for the curator inbox.

jgoley: It may be a good idea to do so. That way we can avoid the logic involved in not showing multi-select and related UI elements.

jgoley:
@nathanathan I moved the extract-incidents table into its own template and also made a small update which prevents the tabs and table from being shown if there are no accepted IRs.

jgoley:
Actually, I forgot to make a change to showing the edit IR modal. I'll fix ASAP.

jgoley:
c51bcc3 Makes the extracted-incidents table completely independent.

PR #307, opened 2017-03-15
allow users to associate with event

Broham:


client/controllers/incidentReports/incidentTable.coffee
```diff
@@ -67,6 +67,16 @@ Template.incidentTable.events
     changeIncidentStatus(false, instance)
     event.currentTarget.blur()
 
+  'click .add': (event, instance) ->
+    instance.data.incidents.find({selected: true}).forEach (incident) ->
+      incident = _id: incident._id
+      incident.userEventId = Template.instance().data.sourceId.get()._str
```
jgoley: I think we want to offer a list of existing events to choose from, then set the `userEventId` to the ID of the chosen event. This will link the source associated with the IR as the event.

jgoley:
@Broham I've established the UI/UX for event association in 958c0bb and 6fcaab5.
Associating IRs to existing events should be pretty straightforward but adding a new event then associating will require some work. We could hold off for another story and temporarily hide the associated UI elements I added. Or I'm happy to take on that side of the project. I think if we work simultaneously, we won't step on each other's toes too much.

Broham:
Awesome thanks @jgoley!  Yeah lets hold off on creating new events from the IR table.  I'm going to be entering stories into PT later today so I will create one for this story and then we can jump on it next week unless you are raring to go.

jgoley:
Sounds good. Let me know if you run into in problems with the setup.

jgoley:
@nathanathan I think this is ready for review when you get a chance.

nathanathan:
When I tried adding some incident reports to an event, they did not appear on the event page. Also, when IRs are added, I recommend keeping the selection so they can be added to other events afterwards.

nathanathan:
Also, It should be possible to click the "Add to Event" button without selecting anything to associate the post with an event without including any incident reports.

jgoley:
Thanks for the feedback @nathanathan. A few questions:
I suppose we should, when adding IRs to an event, also add the source?
Currently we store only one `userEventId` on the IR. Should we change this so we can associated multiple events with an IR or do you mean associated multiple events to a source?

nathanathan:
> I suppose we should, when adding IRs to an event, also add the source?
Yes.
> Currently we store only one userEventId on the IR. Should we change this so we can associate multiple events with an IR or do you mean associated multiple events to a source?
I think we probably will want to make userEventId into an array of multiple events so an IR can be associated with multiple events, but it might be better to copy IRs when they are added to events so edits to an IR in one event can't affect other events. @toph-allen might have an opinion on this.

toph-allen:
>>I suppose we should, when adding IRs to an event, also add the source?

>Yes.

I actually think we shouldn't do this. Down the road, it would lead to more fiddliness with IRs. For example, if you remove an IR from an Event, does its Source also get removed? Not if there are other IRs with the same Source, and not if it's a Source which has been added with no associated IRs.

In this case, I think that we should — maybe invisibly to the user, but maybe with some visual distinction — have two ways that a Source can be displayed on an Event page:

1. A list of the Sources of all IRs (not duplicated). Not a collection or array — just look up the IR and grab its Source.
2. The Sources added without IRs as an array, like we currently do.

The first point is kind of analogous to how we treat Locations — they used to be a separate entity, but now they're just aggregated from all included IRs.

In general, a principle I think we should follow is, where possible, to follow functional principles with Events. In as many places as possible (e.g. our discussion early about the name) the stuff a user sees in an Event isn't static, it's a product of the extracted Incident Reports. I think this lays a solid foundation for more dynamic stuff in this app in the future, or in future apps (e.g. with things like Smart Events).

>>Currently we store only one userEventId on the IR. Should we change this so we can associate multiple events with an IR or do you mean associated multiple events to a source?

>I think we probably will want to make userEventId into an array of multiple events so an IR can be associated with multiple events, but it might be better to copy IRs when they are added to events instead so edits to an IR in one event can't affect other events. @toph-allen might have an opinion on this.

I agree with making `userEventId` into an array so that you can have an IR be associated with multiple Events.

I'd say *"no"* to copying IRs. Since they're are meant to map one-to-one with the information in a Source, I can't imagine a way to edit IRs where you'd *not* want that to propagate to another Event (in other words, if a user is editing an IR to encode different data that's not in the Source, they're using the app wrong, and that's behavior we should endeavor to prevent with UI design and user training.

If, for instance, we've auto-extracted a bunch of IRs in the corpus, and they're automatically clustered into Auto Events or in Smart Events, and a user makes a correction by adding a location to an IR, we'd want that change to show up elsewhere, rather than a duplicate IR showing almost the same info. (That might also screw with deduplication heuristics.)

I know this could be a problematic experience for users — their saved Events being edited by other users, but I'm fine with that for now. I think there are other ways around it that don't dilute the conceptual simplicity we're going for here — the one to one mapping between info in the text and IRs — and I'm fine just having to warn users about this for now.

jgoley:
In 0437f08 I've implemented:
- Allowing users to explicitly add the source to an event
- Implicitly adding the source to an event when adding IRs (if the source isn't already associated)
In doing so I chose to update the UI—giving curators a separate area for source-event association. I feel like keeping source-event association and IR-event association separate will provide a more straightforward UX.

Should we present which events a source is associated with? And filter those out of the selectable list?

I'd say we should create a separate story for associating IRs with multiple events. Seems like the scope of this PR is already pretty stretched.

toph-allen:
>In 0437f08 I've implemented:
>
>- Allowing users to explicitly add the source to an event
>- Implicitly adding the source to an event when adding IRs (if the source isn't already associated)
In doing so I chose to update the UI—giving curators a separate area for source-event association. I feel like keeping source-event association and IR-event association separate will provide a more straightforward UX.

I'm going to recapitulate to make sure we're on the same page: So, "implicitly adding" here means that it's displayed to the user by dint of it being the source of the IR that's part of the event.

>Should we present which events a source is associated with? And filter those out of the selectable list?

Hmm. I think it'd be best to do a quick walk through of the current UI. I'll be in the office this afternoon; is that good? I think I get what you're saying, but I'll have a much clearer thought about it if I see it in action.

>I'd say we should create a separate story for associating IRs with multiple events. Seems like the scope of this PR is already pretty stretched.

Yeah, I can see that. Again, this afternoon when I do a walk-through we can maybe talk through a few different ways to present the relationships to the user, and the different types of actions they might want to take.

jgoley:
@toph-allen _Implicitly adding_ may have been a bad choice of words. When a curator chooses an IR or group of IRs to associate with an event but the source has not been associated with the event, the source is automatically associated without the user knowing.

I think the list of events could also be a separate story so UI review probably shouldn't hold this up from being merged.

toph-allen:
Yeah, go ahead and merge this and we'll do a UI walkthrough in the afternoon. Any changes can be added as new stories.

jgoley:
@nathanathan Do you think this is okay to merge?

nathanathan:
👍

PR #306, opened 2017-03-14
Curator inbox rejected ir table

jgoley:
[PT story](https://www.pivotaltracker.com/story/show/141167197) to show a table of rejected IRs.

PR #308, opened 2017-03-15
Curator inbox add ir

jgoley:
Builds on #306 to accomplish this [PT Story](https://www.pivotaltracker.com/story/show/141165563) to allow curators to add new events to source.

### DIFF
--- a/client/controllers/curatorInbox/curatorSourceDetails.coffee
+++ b/client/controllers/curatorInbox/curatorSourceDetails.coffee
@@ -148,7 +148,7 @@ Template.curatorSourceDetails.events
   'click span.annotation': (event, instance) ->
     incidentId = $(event.currentTarget).data('incident-id')
     Modal.show 'incidentModal',
-      incident: instance.incidentCollection.findOne(incidentId)
+      incident: Incidents.findOne(incidentId)
       articles: [instance.source.get()]
       edit: true
       updateEvent: false
--- a/client/controllers/curatorInbox/curatorSourceDetails.coffee
+++ b/client/controllers/curatorInbox/curatorSourceDetails.coffee
@@ -24,17 +24,6 @@ _markReviewed = (instance, showNext=true) ->
         resolve()
       , 1200
 
-_getSource = (instance, sourceId) ->
-  source = CuratorSources.findOne
-    _id: sourceId
-  instance.reviewed.set source?.reviewed or false
-  instance.source.set source
-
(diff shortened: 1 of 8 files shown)


nathanathan:
When I try to add incident reports they aren't appearing in the table.

jgoley:
This is a known issue, we started by using a local collection like in the event IR addition. I've started a branch to make the tables reactive: 'reactive-IR-tables' I was hoping to continue with it after the open PRs were merged but could continue the work in this branch.

jgoley:
@nathanathan I've updated this PR so the Incident Report tables use the `Incidents` collection instead of the local `incidentsCollection`.

nathanathan:
I am seeing some bugs when I test out the inbox by loading a few unprocessed posts, editing incidents from them, returning to the the posts I edited previously, trying to edit their IRs again, then loading a few more posts, and trying to edit their IRs. Sometimes the IRs from the wrong post show up. Sometimes when I try to save a IR nothing happens. Sometimes I'll get this console error: `Uncaught TypeError: Cannot read property 'top' of undefined
    at incidentTable.coffee:32`.

jgoley:
I think b289324 fixes the issues with:
- IRs unlreated to source showing
- editing issues
- unaccepted IRs sometimes showing with accepted

I'm not sure what's causing the typerror you mentioned. Seems like 037404e should fix it? Maybe you have IRs in the database with the new data schema—IRs with the `annotations` property?

client/controllers/curatorInbox/curatorSourceDetails.coffee
```diff
@@ -162,7 +148,10 @@ Template.curatorSourceDetails.events
   'click span.annotation': (event, instance) ->
     incidentId = $(event.currentTarget).data('incident-id')
```
jgoley: @Broham Yes and no. I think the previous case was also storing wether the IR was accepted or not in the DOM (determined by whether the element has a class, not in the data layer). I'm not sure we can avoid storing the ID in the DOM here because the context of the annotation is the source, not the IR. (But, since we fetch the collection while parsing the text to add the annotations, when the collection changes, the state of the annotations are reactively updated. So the state is still in the data layer). I think the main reason we are storing the ID here is so we can easily access the spans with jQuery (scrollTo) and if we need to access the IR on click (e.g. editing the IR).

Broham: Ahh ok - that makes sense.  I wouldn't worry about this then - looks good. :+1:

PR #309, opened 2017-03-15
Highlight all annotated incident report text with yellow color

jgoley:
The `background` applied to each type were overriding the viewing background given with the `viewing` class.


nathanathan:
:+1:

PR #310, opened 2017-03-16
Open edit IR modal when user clicks source text annotation

jgoley:
[PT Story](https://www.pivotaltracker.com/story/show/141776231)


nathanathan:
ObjectIDs are showing up in the source url field.

nathanathan:
👍

PR #311, opened 2017-03-17
Add text spans to IRs

jgoley:
[PT Chore](https://www.pivotaltracker.com/story/show/141873783)
I wasn't sure if we wanted to only store the case related offsets/text or all related text snippets. I went ahead and started to store all them—seems like doing so will help us show the snippet and annotate the related text (location, date, disease).
Here's the schema:
```
incident: {
...
  annotations: {
    case:
      text: 'Text String',
      textOffSets: [
        [23, 454] 
      ]
    ... (location, date, disease)
  }
}
```


imports/ui/annotation.coffee
```diff
@@ -3,8 +3,9 @@ module.exports =
     lastEnd = 0
     html = ''
     incidents.map (incident)->
-      if incident.textOffsets
-        [start, end] = incident.textOffsets or incident.countAnnotation.textOffsets[0]
+      textOffsets = incident.annotations.case.textOffsets[0][0]
```
nathanathan: This causes errors on incidents that do not have annotations.

jgoley:
How does 2f102f2 look for a fix to the issue with IRs not having annotations.

nathanathan:
That's a good fix. I just noticed the annotations._.textOffsets arrays are 3 levels deep E.g. [[[start, end]]]. Is there a reason we need that or could we bring it down to two?

jgoley:
Is there a chance that an IR will have multiple extracted strings per type? I think that's why I did.

jgoley:
Oh I see the issue. I started fixing this late in the day. Will finish up tomorrow morning.
I think this would be better?:

```json
{
   "case":[
      {
         "textOffsets":[
            573,
            657
         ],
         "text":"under Bisra block of Sundargarh district with the death of 2 villagers"
      }
   ],
   ...
}
```
In the previous version, the case prop was different from the others.

nathanathan:
The textOffsets array in your schema looks like it's only one level deep. I think having two levels would make sense because there could be multiple spans for locations and although there will only be one span for the count, keeping the annotation schema consistent might make the code easier to reason about.

jgoley:
It is but the textOffsets are part of an object in an array. 
So multiple location-related annotations will look like this:
```json
{
  "location":[
    {
      "textOffsets":[
        2539,
        2555
      ]
    },
    {
      "textOffsets":[
        2648,
        2664
      ]
    }
  ]
}
```
In case we want to, at some point, store the text string with the offsets and to stay consistent with the case schema above.

nathanathan:
That sounds good to me.

nathanathan:
👍

PR #312, opened 2017-03-21
Make incident report disease fields use values selected from disease ontology

nathanathan:
This also updates the database to store disease ontology ids and labels instead of free-text values. I am leaving the free-text disease field in the database for now incase we find problems with the automatic mappings of free-text names to disease ontology names.

### DIFF
--- a/client/views/select2.jade
+++ b/client/views/select2.jade
@@ -1,6 +1,7 @@
 template(name="select2")
-  select.form-control.select2(
-    id=selectId
-    name=name
-    multiple=multiple
-    required=required)
+  span
+    select.form-control.select2(
+      id=selectId
+      name=name
+      multiple=multiple
+      required=required)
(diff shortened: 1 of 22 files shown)


client/views/select2.jade
```diff
@@ -0,0 +1,6 @@
+template(name="select2")
+  select.form-control.select2(
+    id=selectId
+    name=name
+    multiple=multiple
+    required=required)
```
jgoley: We should probably show a placeholder? I think in order for it to show, an empty `option` element has to be given?

jgoley:
I'm running into an issue with saving new incident reports. Looks to be due to the submitted incident having no locations but the location field is not marked as invalid.

Should we also switch to the select2 input for creating/editing regular user events?

nathanathan:
I removed the user event disease field. I fixed the other issues you mentioned.

PR #314, opened 2017-03-23
Add incident from selection

jgoley:
This is the first iteration of selecting a span in source text then creating an IR from that span.

**Notes:**
- Since we have to layer elements containing the annotated source text and source text for selection, clicking an annotation within the source text to edit is no longer possible. Later we could implement something like we did in Tater, I suppose? Or we could allow the user to enter a _Selection Mode_ in which the existing annotation highlighting will be reduced and users are allowed to make selections and create IRs. I prefer this idea. We could then control when and in what context IRs are added.
- It's also not currently smart—no properties are extracted from the selected span.
- I went down the road of using the new templates in the suggested IR area but decided to revert in b490c53 because of the click issues mentioned above and we are still using the local collection for IRs within this context of IR acceptance/rejection.

Also: There's a weird bug causing the text of the _add incident_ button to become blurry after appearing. Not sure what's causing it. I've read it could be related to the transitioned `transform` CSS property but all suggested solutions didn't help. I'll keep looking into it.

[PT Story](https://www.pivotaltracker.com/story/show/141873727)


nathanathan:
There is a bug causing highlighted spans to be repeated twice. This should show the suggested incident report modal with the highlighted article snippet so it's possible to verify the correct span of text is being annotated.

jgoley:
@nathanathan Looks like the bug is in master? The incidents appear twice after the source is processed but when the page is refreshed, they appear normally?

For clarification: We should not allow curators to select spans of text containing other IRs?

nathanathan:
I think we should allow spans containing other IR counts to be selected. That could be useful when we re-use the annotation code for geonames where only part of the full name is automatically highlighted.

jgoley:
Sounds good.
6d4561e shows the suggested incident modal. (Thought for the future feature: allow users to select and associated spans related to disease/location/datetime within the suggested IR modal?)
Also, I've fixed the duplicate IRs/annotations bug in #316 since it is in master.

nathanathan:
I don't think we need controls for users to select associated spans for locations/dates/etc. It require a lot of extra clicking and in most cases we would be able to infer the spans from the information the user enters in the form. To be clear, we do want the location/date/disease spans that were automatically associated with the count the users selects to be highlighted.

jgoley:
@nathanathan Currently, though, the only information we have about the IR from selection is the textOffsets (with the exception of the suggested locations). Maybe in a future story we can extract other information from the user selection and the surrounding text?

client/controllers/incidentReports/newIncidentFromSelection.coffee
```diff
@@ -59,11 +60,14 @@ Template.newIncidentFromSelection.helpers
     _getAnnotationData(instance.selection).text
```
jgoley: We may not need to create a function for this since its only called once after your update?

server/methods.coffee
```diff
@@ -128,192 +161,19 @@ Meteor.methods
...
+    enhancements = source.enhancements
+    check enhancements, Object
+    options.url = cleanUrl(source.url)
+    options.publishDate = source.publishDate
+    incidents = createIncidentReportsFromEnhancements(enhancements, options)
+    incidents = incidents.map (incident) ->
+      incident = _.pick(incident, incidentReportSchema.objectKeys())
+    # Remove prior unaccepted incident reports for the article
+    Incidents.remove(
+      url: $regex: regexEscape(options.url) + "$"
```
Broham: Could someone remove unexpected incidents by passing in a URL other than what is expected?  For example, if `promedmail.org` was passed in, would this just remove all unaccepted, autogenerated incidents?

client/controllers/incidentReports/annotatedContent.coffee
```diff
@@ -0,0 +1,46 @@
...
+
+_setSelectingState = (instance, state) ->
+  instance.selecting.set(state)
+
+Template.annotatedContent.onCreated ->
+  @selecting = new ReactiveVar(false)
+  @scrolled = new ReactiveVar(false)
+
+Template.annotatedContent.onRendered ->
+  $sourceContainer = $(@data.relatedElements.sourceContainer)
```
jgoley: True, will remove.

client/controllers/incidentReports/newIncidentFromSelection.coffee
```diff
@@ -0,0 +1,79 @@
...
+  textOffsets = [range.startOffset, range.endOffset]
+  content = selection.anchorNode.textContent
+
+  textOffsets: textOffsets
+  text: content.slice(textOffsets[0], textOffsets[1])
+
+Template.newIncidentFromSelection.onCreated ->
+  @selection = window.getSelection()
+  @selecting = @data.selecting
+  @selecting.set(true)
```
jgoley: Sorry about these. I'll be more rigorous in looking for these sorts of references after refactoring.

server/methods.coffee
nathanathan: promedmail.org wouldn't work but passing in an unexpected suffix could cause problems.

server/methods.coffee
```diff
@@ -178,9 +178,13 @@ Meteor.methods
     incidents = createIncidentReportsFromEnhancements(enhancements, options)
     incidents = incidents.map (incident) ->
       incident = _.pick(incident, incidentReportSchema.objectKeys())
+    # check for unexpected urls
+    if not options.url.startsWith("promedmail.org/post/")
```
Broham: Will this fix the security issue?  Could a user still just pass in: `promedmail.org/post/` and nothing else to accomplish the same thing?  Maybe we should check to confirm there is a 5 digit number at the end of the url?

nathanathan: That wouldn't match anything because the $ requires the match to go to the end of the field.

Broham: Ahh that's right - cool looks good :+1:

PR #318, opened 2017-03-29
Move rejected incidents

Broham:


jgoley:
@Broham I made some fairly extensive updates to the UI.
There's one case that's bugging me — when the accepted IR table does not contain enough IRs to force the rejected IR table to fill the vertical height. I might come back to this and scale those containers dynamically based on the viewport height.

nathanathan:
👍 Looks good!

PR #319, opened 2017-03-29
Curator inbox click ir row to select

jgoley:
[PT Story](https://www.pivotaltracker.com/story/show/142640157)
Not really happy with the UX of editing but this is a temporary solution.


nathanathan:
Could you make it so the edit icons are always visible? It can be helpful when training users to be able to say "click on the pencil icon," and know they'll definitely see it. Otherwise this looks good to me.

jgoley:
@nathanathan ✏️

nathanathan:
👍

PR #320, opened 2017-03-30
Show disease name in IR tables with count

jgoley:
Shows counts and disease like so:
screen shot 2017-03-30 at 5 55 01 pm

[PT story](https://www.pivotaltracker.com/story/show/142404863)


imports/ui/helpers.coffee
```diff
@@ -11,8 +11,8 @@ UI.registerHelper 'formatDateRange', (dateRange)->
   return formatDateRange(dateRange)
 
 UI.registerHelper 'incidentToText', (incident) ->
-  if @cases
-    incidentDescription = pluralize("case", @cases)
+  if incident.cases
```
nathanathan: Why does the variable for the cases property need to change, but not the variable for the deaths and specify properties?

server/methods.coffee
```diff
@@ -177,6 +177,7 @@ Meteor.methods
     options.publishDate = source.publishDate
     incidents = createIncidentReportsFromEnhancements(enhancements, options)
     incidents = incidents.map (incident) ->
+      console.log incident
```
nathanathan: Could you remove this?

imports/ui/helpers.coffee
jgoley: Oh that must have been incidental.
I also didn't realize we already had a helper registered. I'll combine `incidentTypeWithCountAndDisease` with this.

nathanathan:
👍

PR #321, opened 2017-03-31
Misc. bug fixes

nathanathan:
- Annotations property was lost when editing incident in incidentModal
- When no geonames were found querying the geoname lookup services caused an error
- addedByUser properties were not added to autogenered incidents
- modifiedByUser properties are now added when an user edits an incident


jgoley:
👍 
Does the extend solve the annotations property issue?

nathanathan:
@jgoley Yes, it will preserve all the properties that aren't read from the form.

PR #322, opened 2017-03-31
adding spaces and semicolons

Broham:


jgoley:
Maybe since we do this twice now ([another time here](https://github.com/ecohealthalliance/eidr-connect/blob/master/client/controllers/smartEvents/smartEventSummary.coffee#L30-L34)) we should add a global helper?

Broham:
Yeah, that's not a bad point @jgoley - i'll look into that

nathanathan:
I'm going to go ahead and merge this. The other function is a little different (locations are formatted) and they're such short functions a global one won't significantly reduce the amount of code.

PR #324, opened 2017-04-03
Fix IR details - use import instead of require

jgoley:
[PT Bug](https://www.pivotaltracker.com/story/show/142947695)


imports/ui/helpers.coffee
```diff
@@ -1,6 +1,5 @@
 formatLocation = require '/imports/formatLocation.coffee'
```
nathanathan: Can we switch this require call as well?

jgoley: Yep. I've made this change. I wonder if we should work towards a consistent standard—maybe update to the ES6 pattern as we make changes?

nathanathan: That sounds good to me

nathanathan:
👍

PR #325, opened 2017-04-03
Update wording

jgoley:
**Changes**:
- "source" to "document"
- "article" to "document"
- "incident report" to "incident"
- "tracked events" to "curated events"

[PT Story](https://www.pivotaltracker.com/story/show/142864919)

### DIFF
--- a/client/controllers/articles.coffee
+++ b/client/controllers/articles.coffee
@@ -78,7 +78,7 @@ Template.articles.helpers
 
   searchSettings: ->
     id: 'sourceFilter'
-    placeholder: 'Search sources'
+    placeholder: 'Search documents'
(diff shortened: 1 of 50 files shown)


nathanathan:
I see a few places where the wording still needs to be updated. On the curated events table, we should switch "article count" to document count. The article count on the event pages is using the label "sources." The header for the add document modal is currently "Add event source." When no document is selected, the message is "Please select a source." In the curator inbox, there is an "Add source to event" button.

jgoley:
@nathanathan Sorry about missing those occurrences. I looked through again with a much finer toothed comb.

nathanathan:
👍

PR #326, opened 2017-04-04
Add link to smart events

jgoley:
This combines "Curated" and "Smart" event tables into a tabbed UI.
screen shot 2017-04-04 at 12 36 48 pm
It also updates the routing:
`/events` (goes to `/events/curated-events/`
`/events/curated-events/`
`/events/smart-events/`
`/events/curated-events/:_id` (as opposed to `/user-event/:_id`)
`/events/smart-events/:_id`


nathanathan:
There is a bug where if the curated event table is viewed first the document count column is visible on the smart event table, and if the smart event table is viewed first the document count column is not visible on the curated event table.

I'm not sure if it was caused by this PR, but the event-map event links are broken.

jgoley:
@nathanathan In 66055b2 I've fixed the issues with the `Document Count` row. I'm not experiencing issues with the event map route so maybe that was fixed with the updates to route references?

nathanathan:
👍

PR #327, opened 2017-04-04
Scatter bounds

Broham:
Makes it so you can't zoom below 0 and no longer displays decimals (whole numbers only)

### DIFF
--- a/client/controllers/incidentReports/incidentReports.coffee
+++ b/client/controllers/incidentReports/incidentReports.coffee
@@ -84,6 +84,8 @@ Template.incidentReports.onRendered ->
     filters:
       notCumulative: @filters.notCumulative
     # group events
+    minimumZoom:
+      y: 0
     group:
       # methods to be applied when a new group is created
       onEnter: () ->
--- a/imports/charts/Axes.coffee
+++ b/imports/charts/Axes.coffee
@@ -110,13 +110,11 @@ class Axes
     # yAxis
     @yAxis = d3.axisLeft()
       .scale(@yScale)
-
     # yGroup
     @yGroup = @plot.container.append('g')
       .attr('class', 'y scatterPlot-axis')
       .attr('transform', "translate(#{@plot.margins.left}, 0)")
       .call(@yAxis)
-
     # xLabel
     if !@xLabel
       padding = 0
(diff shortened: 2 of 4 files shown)


jgoley:
Should we set a default minimum zoom of 0?
Not sure if this was a part of this story but I recall discussion about limiting the zoom depth—so a user can't zoom infinitely. It's a bit weird now that we round the value of the X axis. Maybe that could be a story of its own? Or maybe we just let the user run down that rabbit hole?
Looks good otherwise.

Broham:
I had thought about defaulting it to 0, but thought that might be weird if someone is using the charting stuff somewhere else in the app where it would make sense to zoom below 0.  I figure we can just leave it with it's current default functionality, but if someone decides they want to limit it they have the option to.

We could limit the zoom, but I figured the rounding thing would be a quick fix for a use case that probably isn't going to be hit very often.  If we set a max zoom we will need to notify the user in some way that they have hit the max zoom level and that's why the chart isn't updating.  If we think it's worth the investment of time we can totally do it, but it seemed like it might be over engineering for a feature that might not be used very often (will users really zoom in 5-6 times an expect meaningful results?).  Thoughts?

Broham:
@nathanathan - do you have any opinion on max zoom vs only showing integers?

nathanathan:
I don't think we need a zoom limit for the reasons you mentioned. Who requested that feature? We should bring them into the discussion if they're not already in it.
I think showing repeated integer ticks is stranger than showing fractional counts at a high zoom levels, and if we adapt this plot to show case count rates we might want fractional values.

Broham:
This was a feature that @toph-allen had requested.  So you are thinking that we shouldn't alter the zoom functionality at all when it comes to showing decimal values?

nathanathan:
Yeah, I think it's okay the way it is since the decimals values don't appear until the user zooms to an unusually high level.

toph-allen:
Yeah, I requested the zoom limits. After reading the discussion here, I don't think it's worth implementing, especially if there's a "Reset Zoom" button so they can easily get back to meaningful zooms. Certainly limiting the Y-axis to > 0 is important, though, to avoid kind of jarring visual weirdness.

toph-allen:
I don't think that truncating or rounding the minor axis ticks makes sense — they should display the value they represent. How about just not displaying minor ticks.

Broham:
Cool - I will yank the decimal stuff I added on zoom and keep the piece that keeps the zoom above 0.  Thanks Toph!

Broham:
Ok - updated.  Should be ready for merge whenever you are ready @nathanathan

PR #328, opened 2017-04-04
Curator inbox click annotation to scroll

jgoley:
[PT Story](https://www.pivotaltracker.com/story/show/142863589)
There's one UX-related issue that may not be resolvable—we can't change the cursor when the mouse is over the annotations. It's not a huge deal but it would be nice to have some sort of indication, especially since we plan to add additional features post-click.


nathanathan:
👍 Looks good!

PR #332, opened 2017-04-05
Format dates within curator inbox incident tables

jgoley:
[PT Story](https://www.pivotaltracker.com/story/show/142809815)


nathanathan:
For these dates I think it would be better to use a single column and the date formatting function from the event page incident table. The reason for using two columns before was just to make it so the table could be downloaded as an easy to parse CSV when it was used on the extract-incidents page.

jgoley:
Just had a thought.
Maybe we can update the formatDateRange helper to only inlcude the year at the end of the string if the range is within the same year. Like: `March 1 - March 30, 2017`?

nathanathan:
👍 Looks good

PR #333, opened 2017-04-06
Replace CuratorSources with Articles, prevent duplicate grits-api calls

nathanathan:
This replaces the CuratorSources collection with the Articles collection. The articles that were stored in the curator sources collection are copied into the Articles collection so their "reviewed" status is not lost.

This also adds a getArticleEnhancementsAndUpdate method that prevents users from repeatedly sending the same article to the grits API.

@Broham and @jgoley could you both take a look at this PR? It needs a thorough review.


server/cleanUrls.coffee
```diff
@@ -81,16 +80,15 @@ updateMetadataLinks = (collection, host, domain) ->
...
     host = 'promedmail'
     domain = 'org'
     updateUrls(Articles, host, domain)
     updateUrls(IncidentReports, host, domain)
-    updateMetadataLinks(CuratorSources, host, domain)
 
-if Meteor.isServer
-  Meteor.startup ->
-    cleanPromedUrls()
+# if Meteor.isServer
```
Broham: did you mean to leave this commented out?  If so - should we just delete it?

server/methods.coffee
```diff
@@ -75,6 +75,35 @@ Meteor.methods
...
+  # and create incidents in the database.
+  getArticleEnhancementsAndUpdate: (article) ->
+    dbArticle = Articles.findOne(_id: article._id)
+    if not dbArticle
+      throw Meteor.Error('invalid-article')
+    if dbArticle.enhancements
+      if dbArticle.enhancements.processingStartedAt
+        # If the processing started less than 100 seconds ago do not resubmit
+        # the aritcle.
+        console.log 't', new Date() - dbArticle.enhancements.processingStartedAt
```
Broham: should this be removed?

jgoley:
Looks good! 👍

server/startup.coffee
```diff
@@ -99,3 +93,18 @@ Meteor.startup ->
             resolvedDisease:
               id: "userSpecifiedDisease:#{incident.disease}"
               text: "Other Disease: #{incident.disease}"
+
+Meteor.startup ->
```
jgoley: Is this necessary? We call `startup` on line 10.

PR #336, opened 2017-04-06
Remove IR from selected collection when accepted

jgoley:
[PT Bug](https://www.pivotaltracker.com/story/show/143263801)


nathanathan:
👍

PR #340, opened 2017-04-10
Prevent add IR btn from switching locations immediately

jgoley:
This also reduces the vertical movement of the text due to the addition of the button at the top of the text container.
[PT Bug](https://www.pivotaltracker.com/story/show/143328867)


nathanathan:
👍

PR #342, opened 2017-04-10
Only show plot zoome reset btn when plot zoomed

jgoley:
[PT Bug](https://www.pivotaltracker.com/story/show/143333009)


nathanathan:
👍

PR #343, opened 2017-04-10
Curator inbox sort incidents after addition

jgoley:
I suppose minimongo can't handle sorting like this: `sort: 'annotations.case.0.textOffsets.0': 1` I replaced it with `_.sortBy` to fix the bug where newly added Incidents were ending up at the bottom of the list.
I also noticed we, at some point, started to set the incidents loaded flag before the subscription is ready so I fixed that.
[PT Bug](https://www.pivotaltracker.com/story/show/143314903)

### DIFF
--- a/client/controllers/curatorInbox/curatorSourceDetails.coffee
+++ b/client/controllers/curatorInbox/curatorSourceDetails.coffee
@@ -69,16 +69,17 @@ Template.curatorSourceDetails.onRendered ->
           $title.tooltip('hide').attr('data-original-title', '')
         else
           $title.attr('data-original-title', title)
-      @subscribe 'ArticleIncidentReports', source.url
-      if source.enhancements?.dateOfDiagnosis
-        instance.incidentsLoaded.set(true)
-      else
-        Meteor.call 'getArticleEnhancementsAndUpdate', source, (error, enhancements)=>
-          if error
-            notify('error', error.reason)
-          else
-            source.enhancements = enhancements
+      @subscribe 'ArticleIncidentReports', source.url,
+        onReady: ->
+          if source.enhancements?.dateOfDiagnosis
             instance.incidentsLoaded.set(true)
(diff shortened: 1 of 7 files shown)


client/controllers/curatorInbox/curatorSourceDetails.coffee
```diff
@@ -69,16 +69,17 @@ Template.curatorSourceDetails.onRendered ->
...
-      if source.enhancements?.dateOfDiagnosis
-        instance.incidentsLoaded.set(true)
-      else
-        Meteor.call 'getArticleEnhancementsAndUpdate', source, (error, enhancements)=>
-          if error
-            notify('error', error.reason)
-          else
-            source.enhancements = enhancements
+      @subscribe 'ArticleIncidentReports', source.url,
+        onReady: ->
```
nathanathan: There is a bug in this PR where some of the incidents aren't loaded into the IR table. I suspect it is related to the onReady callback here.

jgoley: I'm having trouble replicating this. Seems like both the annotated document and the tables will both render after the subscription is ready, so the incidents in either should be consistent?

I did notice a bug with calling the jquery object of the instance before the template is rendered which is fixed in 62ba40d

nathanathan: I tracked down the issue causing incidents not to load. If an incident is created without annotations it will trigger an error in the sort function.

I still think there is a problem with the way the onReady callback is being used here in that it leads to an extra server round trip before article processing can begin.

jgoley: I think I've corrected the bug caused by an incident having no annotations in 0115d98 and removed the `onReady` in bfa7bc3.

nathanathan:
👍

PR #345, opened 2017-04-11
Curator inbox switch between document groups

jgoley:
[PT Story](https://www.pivotaltracker.com/story/show/143340613)

This assumes we distinguish manually added documents from machine-added by whether it has a `addedByUserId` ([here](https://github.com/ecohealthalliance/eidr-connect/blob/0e81e8651d93b07296bb3a6d371a87015e7c3a9f/client/controllers/curatorInbox/curatorInbox.coffee#L83)) This correct?
The way users select the feed may be too ambiguous. Thoughts?

Also: I figured I'd go ahead and add Promed as a feed in implementing this story.

In implementing I discovered [this bug](https://www.pivotaltracker.com/story/show/143522211)


server/startup.coffee
```diff
@@ -94,6 +96,23 @@ Meteor.startup ->
...
+    promedFeed =
+      title: 'ProMED-mail'
+      url: 'promedmail.org/post/'
+      addedDate: new Date()
+    feedSchema.validate(promedFeed)
+    promedFeedId = Feeds.insert(promedFeed)
+
+  Articles.find(
+    url: $regex: /promedmail.org/
+    feed: $exists: false
```
nathanathan: Should this be true?

nathanathan:
> This assumes we distinguish manually added documents from machine-added by whether it has a addedByUserId (here) This correct?

Yes

> The way users select the feed may be too ambiguous. Thoughts?

I thought it was pretty clear.

For the user feeds we I think it would be better to sort/search by the added date rather than the publish date.

The feed selector appears to have two empty options.

server/startup.coffee
jgoley: Good point. I suppose we need to update it to `feedId`

jgoley:
Thanks for the feedback @nathanathan.
I'll go ahead and sort the user added results. 

The empty options in the feed selector are, I assume, a few feeds added in the past. The old data is lacking a `title` so maybe we should show the feed url for existing feeds?

server/startup.coffee
```diff
@@ -94,6 +96,26 @@ Meteor.startup ->
               id: "userSpecifiedDisease:#{incident.disease}"
               text: "Other Disease: #{incident.disease}"
 
+  promedFeed = Feeds.findOne(url: $regex: /promedmail.org/)
+  if not promedFeed?.title
+    newFeedProps =
+      title: 'ProMED-mail'
+      url: 'promedmail.org/post/'
+    feedSchema.validate(newFeedProps)
+    Feeds.upsert promedFeed?._id,
```
nathanathan: If there is no promeFeed I think this would end up inserting it with a null id.

nathanathan:
Showing the url for existing feeds sounds good.

I noticed one other issue, ProMED posts that are added by users aren't showing up in the user added feed.

server/startup.coffee
jgoley: Seems like if the id is null, then a new document would be inserted since its an `upsert`?

jgoley:
_(mentioning for posterity)_
Nathan and I discussed via Slack the bug he mentioned about posts not showing in the user added feed. He was testing by selecting a suggested source which most likely won't have a `addedByUserId` and won't be caught by the user added query.

server/startup.coffee
nathanathan: You're right. It generates an id too.

server/startup.coffee
```diff
@@ -94,6 +96,26 @@ Meteor.startup ->
               id: "userSpecifiedDisease:#{incident.disease}"
               text: "Other Disease: #{incident.disease}"
 
+  promedFeed = Feeds.findOne(url: $regex: /promedmail.org/)
```
nathanathan: Let's replace these regex's with an exact url for the automatically added ProMED feed. If we add ProMED regional feeds this type of matching could pick up the wrong feed.

nathanathan:
👍 Looks good

PR #351, opened 2017-04-21
Fix tests

jgoley:


jgoley:
This is almost ready. Currently some tests are failing due to failed connections with the Grits api.

frosario:
Good and bad news @jgoley @nathanathan.. We're making progress, but now it's complaining about not being able to connect to spa:
`W20170428-13:39:32.766(0)? (STDERR) Unable to connect to remote SPA mongodb.`
Do we want to connect to spa or stub out the SPA call too?

nathanathan:
🎊 Nice work!

PR #352, opened 2017-04-24
Add autoprocessing

nathanathan:
This automatically submits the most recently added articles to the grits api. It also fixes a bug that prevented uncached historical date ranges from being queried.


Broham:
looks good :+1:

PR #353, opened 2017-04-25
Articles multiple events

Broham:
Allows articles to be assigned to multiple events


methods/articles.coffee
```diff
@@ -2,28 +2,32 @@ Articles = require '/imports/collections/articles.coffee'
...
       # Check if Document is in collection
       sourceQuery = url: $regex: "#{regexEscape(source.url)}$"
       existingSource = Articles.findOne(sourceQuery)
       if existingSource
-        Articles.update sourceQuery,
-          $set: userEventId: source.userEventId
+        #if this source is already in the DB and the userEvent isn't already associated - push it
+        if !existingSource.userEventIds or existingSource.userEventIds.indexOf(source.userEventId) == -1
+          Articles.update sourceQuery,
+            $push: userEventIds: eventId
```
nathanathan: We could skip the prior association check by using $addToSet

nathanathan:
There are some references to the old userEventId property that need to be updated:
```
/eidr-connect/server/methods.coffee:
	130:     Articles.find(userEventId: eventId).forEach (relatedEventSource) ->

/eidr-connect/imports/schemas/article.coffee:
	22:   userEventId:

/eidr-connect/client/controllers/events/createEventModal.coffee:
	42:             userEventId: result.insertedId

/eidr-connect/client/controllers/deleteConfirmation/deleteConfirmationModalBody.coffee:
	17:         Meteor.call 'removeEventSource', id, data.userEventId, (error) ->

/eidr-connect/server/router.coffee:
	63:   userEventId = @request.body.eventId ? ""
	66:   if userEventId.length and article.length
	67:     userEvent = getUserEvents().findOne(userEventId)
	69:       existingArticle = Articles.find({url: article, userEventId: userEventId}).fetch()
	72:         Articles.insert({userEventId: userEventId, url: article})
	89:       $in: _.pluck(articles, 'userEventId')
	95:       userEventId: event._id
```

Broham:
@nathanathan - I pushed an update to address your comments.  FYI - I believe the 3rd and 4th examples you gave of `userEventId` are correct since they have to do with calling methods that use a single `userEventId` rather than interacting with the database.

nathanathan:
In the third one the userEventId doesn't need to be there since addEventSource now expects an eventId parameter instead. I would double check the userEventId parameter on the fourth one to make sure it's not coming from an article object.

nathanathan:
I'm seeing a bug on the event page where if I delete an document then try to add a new one, the new document is not added.

Also, when I try to add documents to more than one event in the inbox, they are only appearing in the first event I add them to.

jgoley:
Looks like there is an issue related to Nathan's comment about docs only appearing on the first event: when adding a document already associated with another event on a different event's page, the user receives confirmation that the document is added but it does not appear in the event's list. I'm also getting an error:
```
Exception while invoking method 'getArticleEnhancements' TypeError: Cannot read property 'url' of null at [object Object].getArticleEnhancements (server/dbUtils.coffee:1:1)
```

Broham:
It appears that in the 4th example is coming from an object that does not represent a document in the database.  It contains information about the current article as well as the current user event, but does not use the naming conventions we are using for documents in either collection.  I think it is ok.  

Also - the error saving docs appears to be coming from my change to the use of `addToSet`.  I've changed it back to `push` with the check in the conditional above and it seems to be working again.  I think I may have misunderstood what you were saying when you suggested i use `addToSet` instead.  Could you go into more detail about why I should be using it here instead of `push`?

jgoley:
Looks like the association is working now. 👍  
I'm not sure the issue I mentioned earlier is really an issue applicable to this PR. Maybe we should check if an article exists before adding the article?

nathanathan:
The problem was the `if !existingSource.userEventIds` condition was preventing the document from being updated when there were userEventIds associated with it rather than the $addToSet function. The advantage of $addToSet over $push is that it removes the need to check for existing ids.

Broham:
I see what you are saying @nathanathan.  Pushed an update.

nathanathan:
I'm still seeing the bug where if I delete a document then try to add a new one the new document is not added.

Broham:
@nathanathan - weird, I'm not able to reproduce this.  Below are the steps i'm taking, do you get the error if you take them?

 1. Go to an event (http://localhost:3000/events/curated-events/QMvHyypQdWYzhYvKR for example)
 2. Add a new document (http://www.promedmail.org/post/753328 for example) and uncheck the "Suggest Incidents" button
 3. Delete the newly added document (titled `PRO> Salmonella, antibiotic resistance - USA`)
 4. Document should be gone.  Re-add using the steps above

When I do this the document appears in the doc list both times.  Are these the same steps you are taking?  If so - when you try this with the same events/articles is the doc still not appearing?

Broham:
I also tested from the inbox and wasn't able to reproduce there

jgoley:
Looks like its working okay for me.
I would suggest making a wording change to the tooltip of the icon/button associated with removing a document from an event since the user isn't deleting the document completely from the system—just removing from the event. So maybe "Remove from Event".
It also may be a good idea to change the icon to http://fontawesome.io/icon/times/

Broham:
Good point Jonathan - I will change the wording.  There is currently no way to remove documents from the system correct?

methods/articles.coffee
```diff
@@ -2,28 +2,31 @@ Articles = require '/imports/collections/articles.coffee'
...
+        Meteor.call("editUserEventArticleCount", eventId, 1)
+        existingSource._id
       else
         if source.url
           insertArticle =
             url: source.url
             title: source.title
-            userEventId: source.userEventId
+            userEventIds: [eventId]
           insertArticle = source
```
nathanathan: It looks like the userEventIds property is being wiped out here. I think this is the reason I was seeing the bug where articles were not being added.

Broham: @nathanathan - that section of code only runs when you are inserting a new article so there shouldn't be a `userEventIds` property to wipe our, correct?  There is an if statement above that that handles updating `userEventIds` on existing sources so that that field isn't wiped out.

Are you still seeing the bug you mentioned earlier?

nathanathan: The insertArticle variable is being overwritten. The userEventIds property doesn't make it into the database on the initial article.

Broham: Yeah looks like we don't even need `insertArticle` - I reworked some of the logic, should be working now.

nathanathan:
Right, there currently is no way to remove documents from the system.

Broham:
@nathanathan - conflicts should be resolved now

nathanathan:
The "I delete an existing document" and "I edit an existing document" tests are failing where they passed before.

server/startup.coffee
```diff
@@ -1,11 +1,16 @@
 incidentReportSchema = require '/imports/schemas/incidentReport.coffee'
 Incidents = require '/imports/collections/incidentReports.coffee'
+Articles = require '/imports/collections/articles.coffee'
 
 Meteor.startup ->
   # Clean-up curatorInboxSourceId when user goes offline
   Meteor.users.find({'status.online': true}).observe
     removed: (user) ->
       Meteor.users.update(user._id, {$set : {'status.curatorInboxSourceId': null}})
+  # update articles to use arrays instead of strings for their UserEventId values
```
nathanathan: I changed the way database updates work. This should go at the end of updaters.coffee and the data version needs to be incremented.

Broham:
@nathanathan - tests appear to be passing using the new data loading methodology.  I also moved the startup script changes I had to `updaters.coffee`

nathanathan:
👍 Looks good!

PR #354, opened 2017-04-27
Add incident CSV download to event page

nathanathan:
Currently the document properties all have placeholder values. After the new articleId property is merged I will rebase this and use the property to add the real values.


jgoley:
Oh, I know this is due to the plugin we use to create CSVs needing a UI table, but it is a little weird that we name the button "Download CSV" then they see a modal and are required to click another button to actually download the CSV file. I suppose we could hide the table and show a modal with some sort of indicator that it has been downloaded? But I can see the benefit of showing the table so if we want to keep it, maybe we automatically download the CSV after the modal appears?

nathanathan:
@jgoley I want the modal dialog to appear before the download starts so we can add controls to it for customizing the exported CSV such as hiding/renaming columns and aggregating counts.

jgoley:
Oh that's a good idea. Maybe we should change the text of the button to "Export" to reduce potential confiusion?

client/views/incidentReports/incidentReports.jade
```diff
@@ -6,7 +6,7 @@ template(name="incidentReports")
           +addIncidentReport
         button.open-download-csv.right-action.btn.btn-primary(type="button")
           i.fa.fa-download
-          span Download CSV
+          span Export CSV
```
jgoley: I think we can also remove "CSV"

jgoley:
With the exception of the wording change, the updates look good 👍

PR #355, opened 2017-04-28
Show warning if document does not contain text content

jgoley:
One of the issues the proMed folks had I think was related to them trying to view and process a document with an invalid URL (http://www.promedmail.org/post/undefined in this case). Not sure how it happened but I suppose we need to check for this on submission? Or just stay with letting it fail gracefully like this PR does.
I think this is just an initial step—seems like we should allow the curator to edit or delete the document? Maybe both?


nathanathan:
There is a story in the backlog for deleting documents. I'll add a schema validation check for articles.

PR #356, opened 2017-05-01
Validate documents on submission

nathanathan:
This adds schema validation for article/documents. It also fixes a bug where unresolved geoname ids would trigger exceptions instead of being ignored, and it might help with memory leak we saw on the server by using setInterval instead of recursive setTimeout calls.

### DIFF
--- a/server/methods.coffee
+++ b/server/methods.coffee
@@ -67,14 +67,15 @@ Meteor.methods
           featureClass: geoname.featureClass
           featureCode: geoname.featureCode
           alternateNames: geoname.alternateNames
-      locationAnnotations = locationAnnotations.filter (loc)->
+      locationAnnotations.forEach (loc)->
         geoname = geonamesById[loc.geoname.geonameid]
         if geoname
           loc.geoname = geoname
-          true
         else
-          console.log "Missing geoname for id: " + loc.geoname.geonameid
-          false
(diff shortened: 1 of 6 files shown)


server/methods.coffee
```diff
@@ -67,14 +67,15 @@ Meteor.methods
           featureClass: geoname.featureClass
           featureCode: geoname.featureCode
           alternateNames: geoname.alternateNames
-      locationAnnotations = locationAnnotations.filter (loc)->
+      locationAnnotations.forEach (loc)->
         geoname = geonamesById[loc.geoname.geonameid]
         if geoname
           loc.geoname = geoname
```
jgoley: Wondering if this is this necessary? Maybe we just do what's in the else block if there's no `geoname`?

nathanathan: It is necessary. The geoname in the grits-api enhancement is different from the one that the geoname lookup service returns. We need the labels administrative division labels from it.

jgoley: Is the lookup version supposed to replace the version in the original `enhancements` object? How is it being replaced?

nathanathan: That's right. The original version is replaced by setting it to the look-up version stored in the geonamesById map.

jgoley:
👍

PR #357, opened 2017-05-01
Sort events by lastModifiedDate

jgoley:
[PT Story](https://www.pivotaltracker.com/story/show/144099257)
Required a little finagling with the way we are managing sorting.


nathanathan:
👍

PR #358, opened 2017-05-02
Optimize startup and disable broken event-map page

nathanathan:
The event-map page needs to be refactored. It is subscribing to too many incidents crashing the server.

The startup optimizations work by asynchronously validating incident reports and only running database update code if a dataVersion property I added indicates the database needs the update.

I also made some modifications to the tests to ignore the scenarios we expect to fail and not try click the hidden enhance button.


jgoley:
Good call with versioning the database.
👍

Broham:
Looks good :+1:

PR #359, opened 2017-05-05
Fix event map

nathanathan:
This makes it so the event map only subscribes to incidents for the currently visible page of events. Subscribing to the full collection was causing the server to crash with memory errors.


jgoley:
**Noticed a few things:**
- Looks like the date filtration is broken?
- `undefined` data appears popups of events with incidents with a type of "other":screen shot 2017-05-05 at 1 23 21 pm
- Maybe when the user clears all the selected events we should return the zoom level of the map to its initial level?
- I also noticed some UI/UX issues like the calendar icon getting a border in its active state and more importantly the variability of the list's vertical height changing the position of the pagination buttons. I'll try to fix them this afternoon but I suppose they could be fixed in another PR.

jgoley:
I'm seeing some incorrect dates at the end of the list:
screen shot 2017-05-08 at 12 03 53 pm
Otherwise it looks good.

PR #360, opened 2017-05-08
Tests incident suggestion

jgoley:
Fixes [this bug](https://www.pivotaltracker.com/story/show/145059173) and creates a Curator Inbox testing feature which tests document addition and and replaces tests for processing.
For some strange reason the tests fail in Chrome but pass in PhantomJS. I suppose the browser crashes or the user is logged out somehow during the waitForExist methods [here](https://github.com/ecohealthalliance/eidr-connect/blob/83b4b95de5a5c671b20e372c151b8402b718e34f/tests/features/step_definitions/curator_inbox_steps.coffee#L46-L50).

### DIFF
--- a/client/controllers/curatorInbox/curatorInbox.coffee
+++ b/client/controllers/curatorInbox/curatorInbox.coffee
@@ -85,7 +85,7 @@ Template.curatorInbox.onRendered ->
 
   @autorun =>
     if @selectedFeedId.get() in getCustomFeedArray()
-      latestSourceDate = moment().format('L')
+      latestSourceDate = moment().endOf('day').format('L')
       range = @dateRange.get()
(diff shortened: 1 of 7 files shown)


nathanathan:
There could be a timing issue in the "I add a new test document..." test that puts the app into the wrong state causing the next step to fail.

jgoley:
Strangely it looks like some of the issues I was having were related to a few console logs. 
I also bumped up the time for the 2 waitForExist method calls to 5 seconds which wait for the document to process. I suppose this will give us enough latitude if and when the  grits server is slower than normal.

jgoley:
@nathanathan I've determined this is not a timing issue. I suppose it might be another filtration bug. I'll investigate further in the morning.

nathanathan:
👍

PR #361, opened 2017-05-08
Fix first document not selected

jgoley:
[Fixes this bug](https://www.pivotaltracker.com/story/show/144609153)
Also fixes another bug where the first user added document was not being selected.


nathanathan:
When I filter by a historical time period the selected document is not in the list on the left. It seems to come from the day after the time period. This might be a timezone related issue.

jgoley:
The issue was related to this line: https://github.com/ecohealthalliance/eidr-connect/blob/master/client/controllers/curatorInbox/curatorInbox.coffee#L107
I unwisely added it to catch newly added documents. I suppose this is a better approach: 74b7fd1

nathanathan:
👍

PR #363, opened 2017-05-15
Update event incident association

jgoley:
[PT Chore](https://www.pivotaltracker.com/story/show/143506429)

Stores incidents on events like so:
```javascript
  "incidents": [
    {
      "id": "QAbNzThLNd2m7iyzu",
      "associationDate": ISODate("2017-05-12T02:30:34.030Z"),
      "associationUserId": "DwM8KgCMHjKxgpRdj"
    },
    {
      "id": "z6sEaZhedATuJz6AE",
      "associationDate": ISODate("2017-05-12T02:30:34.218Z"),
      "associationUserId": "DwM8KgCMHjKxgpRdj"
    }
  ]
```
I also updated the article list on event pages to show articles with implied association—incidents from the article have been associated with the event, but not the article itself (explicitly).

To prevent inconsistent data with counts I removed the document count from the table. Is this crucial?

### DIFF
--- a/client/controllers/events/curatedEvent.coffee
+++ b/client/controllers/events/curatedEvent.coffee
@@ -1,6 +1,7 @@
-Incidents = require '/imports/collections/incidentReports.coffee'
-UserEvents = require '/imports/collections/userEvents.coffee'
-Articles = require '/imports/collections/articles.coffee'
+EventIncidents = require '/imports/collections/eventIncidents'
+EventArticles = require '/imports/collections/eventArticles'
+UserEvents = require '/imports/collections/userEvents'
+
 #Allow multiple modals or the suggested locations list won't show after the loading modal is hidden
 Modal.allowMultiple = true
 
@@ -8,16 +9,14 @@ Template.curatedEvent.onCreated ->
   @editState = new ReactiveVar(false)
   @loaded = new ReactiveVar(false)
   userEventId = @data.userEventId
-  @subscribe "userEvent", @data.userEventId
+
+  @subscribe 'userEvent', @data.userEventId, =>
+    @loaded.set(true)
 
   @autorun =>
     userEvent = UserEvents.findOne(userEventId)
     if userEvent
       document.title = "Eidr-Connect: #{userEvent.eventName}"
-
-  @subscribe "eventArticles", userEventId
-  @subscribe 'eventIncidents', userEventId, =>
-    @loaded.set(true)
 
 Template.curatedEvent.onRendered ->
   new Clipboard '.copy-link'
@@ -27,18 +26,19 @@ Template.curatedEvent.helpers
     UserEvents.findOne(Template.instance().data.userEventId)
 
   eventHasArticles: ->
-    Articles.find().count()
+    EventArticles.find().count()
 
   articleData: ->
-    articles: Articles.find()
-    userEvent: UserEvents.findOne(Template.instance().data.userEventId)
+    instance = Template.instance()
+
+    articles: EventArticles.find()
+    userEvent: UserEvents.findOne(instance.data.userEventId)
 
   incidents: ->
-    event = UserEvents.findOne(Template.instance().data.userEventId)
-    Incidents.find(_id: $in: _.pluck(event.incidents, 'id'))
+    EventIncidents.find()
 
   incidentCount: ->
-    Incidents.find().count()
+    EventIncidents.find().count()
 
   isEditing: ->
     Template.instance().editState.get()
@@ -64,14 +64,14 @@ Template.curatedEvent.helpers
 
     userEvent = UserEvents.findOne(instance.data.userEventId)
     userEvent: userEvent
-    articles: Articles.find()
-    incidents: Incidents.find(_id: $in: _.pluck(userEvent.incidents, 'id'))
+    articles: EventArticles.find()
+    incidents: EventIncidents.find()
 
   loaded: ->
     Template.instance().loaded.get()
 
   documentCount: ->
-    Articles.find().count()
+    EventArticles.find().count()
 
 Template.curatedEvent.events
   'click .edit-link, click #cancel-edit': (event, instance) ->
@@ -79,7 +79,7 @@ Template.curatedEvent.events
 
   'click .open-incident-form-in-details': (event, instance) ->
     Modal.show 'incidentModal',
-      articles: Articles.find()
+      articles: EventArticles.find()
       userEventId: instance.data.userEventId
       add: true
 
--- a/client/controllers/events/curatedEvent.coffee
+++ b/client/controllers/events/curatedEvent.coffee
@@ -7,15 +7,23 @@ Modal.allowMultiple = true
 Template.curatedEvent.onCreated ->
   @editState = new ReactiveVar(false)
   @loaded = new ReactiveVar(false)
+  @incidentIds = new ReactiveVar([])
   userEventId = @data.userEventId
-  @subscribe "userEvent", @data.userEventId, =>
+  @subscribe "userEvent", @data.userEventId
+
+  @autorun =>
     userEvent = UserEvents.findOne(userEventId)
-    document.title += ": #{userEvent.eventName}"
-    incidentIds = _.map userEvent.incidents, (incident) ->
-      incident.id
+    if userEvent
+      document.title += ": #{userEvent.eventName}"
+      @incidentIds.set _.map userEvent.incidents, (incident) ->
+        incident.id
+
+  @autorun =>
     @subscribe "eventArticles", userEventId
-    @subscribe 'eventIncidents', incidentIds, =>
-      @loaded.set(true)
+    incidentIds = @incidentIds.get()
+    if incidentIds
+      @subscribe 'eventIncidents', incidentIds, =>
+        @loaded.set(true)
 
 Template.curatedEvent.onRendered ->
   new Clipboard '.copy-link'
@@ -68,12 +76,12 @@ Template.curatedEvent.events
   'click .open-incident-form-in-details': (event, instance) ->
     data = instance.data
     Modal.show 'incidentModal',
-      articles: data.articles
-      userEventId: data.userEvent._id
+      articles: Articles.find()
+      userEventId: UserEvents.findOne(instance.data.userEventId)
       add: true
 
   'click .open-source-form-in-details': (event, instance) ->
-    Modal.show('sourceModal', userEventId: instance.data.userEvent._id)
+    Modal.show('sourceModal', userEventId: instance.userEventId)
 
   'click .tabs li a': (event) ->
     event.currentTarget.blur()
--- a/client/controllers/events/curatedEvent.coffee
+++ b/client/controllers/events/curatedEvent.coffee
@@ -84,7 +84,7 @@ Template.curatedEvent.events
       add: true
 
   'click .open-source-form-in-details': (event, instance) ->
-    Modal.show('sourceModal', userEventId: instance.userEventId)
+    Modal.show('sourceModal', userEventId: instance.data.userEventId)
 
   'click .tabs li a': (event) ->
     event.currentTarget.blur()
--- a/client/controllers/events/curatedEvent.coffee
+++ b/client/controllers/events/curatedEvent.coffee
@@ -7,7 +7,7 @@ Modal.allowMultiple = true
 Template.curatedEvent.onCreated ->
   @editState = new ReactiveVar(false)
   @loaded = new ReactiveVar(false)
-  @incidentIds = new ReactiveVar([])
+  @incidentIds = new Meteor.Collection(null)
   userEventId = @data.userEventId
   @subscribe "userEvent", @data.userEventId
 
@@ -15,11 +15,12 @@ Template.curatedEvent.onCreated ->
     userEvent = UserEvents.findOne(userEventId)
     if userEvent
       document.title = "Eidr-Connect: #{userEvent.eventName}"
-      @incidentIds.set _.map userEvent.incidents, (incident) ->
-        incident.id
+      userEvent.incidents.forEach (incident) =>
+        @incidentIds.upsert id: incident.id,
+          id: incident.id
 
   @autorun =>
-    incidentIds = @incidentIds.get()
+    incidentIds = _.pluck(@incidentIds.find().fetch(), 'id')
     @subscribe "eventArticles", userEventId, incidentIds
     if incidentIds
       @subscribe 'eventIncidents', incidentIds, =>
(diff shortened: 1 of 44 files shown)


client/controllers/events/curatedEvent.coffee
```diff
@@ -1,15 +1,47 @@
...
+  @editState = new ReactiveVar(false)
+  @loaded = new ReactiveVar(false)
+  @incidentIds = new ReactiveVar([])
+  userEventId = @data.userEventId
+  @subscribe "userEvent", @data.userEventId
+
+  @autorun =>
+    userEvent = UserEvents.findOne(userEventId)
+    if userEvent
+      document.title += ": #{userEvent.eventName}"
```
nathanathan: If the event is updated this could run again and mess up the title.

client/controllers/maps/eventMap.coffee
```diff
@@ -73,60 +74,73 @@ Template.eventMap.onRendered ->
...
-        padding: [20, 20]
+    templateEvents = @templateEvents.get()
+
+    # Create an array of the ids of the incidents of the selected events
+    incidentIds = []
+    @selectedEvents.find().fetch().forEach (event) ->
+      incidentIds = _.union(incidentIds, _.pluck(event.incidents, 'id'))
+    incidentIds
+
+    @subscribe 'mapIncidents', incidentIds, =>
```
nathanathan: I think the way the subscription was done before was better because it didn't have to make a trip to the server every time an event is selected.

methods/incidentReports.coffee
```diff
@@ -15,11 +15,23 @@ Meteor.methods
     incident.addedByUserName = user.profile.name
     incident.addedDate = new Date()
     newId = Incidents.insert(incident)
-    if incident.userEventId
-      Meteor.call("editUserEventLastModified", incident.userEventId)
-      Meteor.call("editUserEventLastIncidentDate", incident.userEventId)
+    if userEventId
+      Meteor.call('addIncidentReportToEvent', userEventId, newId)
+      Meteor.call("editUserEventLastModified", userEventId)
```
nathanathan: The later two calls should be moved to addIncidentReportToEvent.

methods/incidentReports.coffee
```diff
@@ -35,18 +47,20 @@ Meteor.methods
       Meteor.call("editUserEventLastIncidentDate", incident.userEventId)
     return incident._id
 
-  editIncidentReport: (incident) ->
+  editIncidentReport: (incident, userEventId) ->
```
nathanathan: I don't think the userEventId parameter should be added here. I don't see any updated calls to this method, and adding incidents to events would be better to do through the "addIncidentReportToEvent" method.

nathanathan:
Removing the document count seems reasonable. Could you replace it with the incident count? Also, I think the code that updates the article counts hasn't been removed yet.

client/controllers/events/curatedEvent.coffee
```diff
@@ -40,12 +79,12 @@ Template.curatedEvent.events
   'click .open-incident-form-in-details': (event, instance) ->
     data = instance.data
     Modal.show 'incidentModal',
-      articles: data.articles
-      userEventId: data.userEvent._id
+      articles: Articles.find()
+      userEventId: UserEvents.findOne(instance.data.userEventId)
```
jgoley: I think we can remove this? Looks to be old code and replaced by this: https://github.com/ecohealthalliance/eidr-connect/blob/ac8eeff38ee27022b06db37ceed3b2bf1748f9e1/client/controllers/incidentReports/addIncidentReport.coffee#L2-L7

jgoley: Oh never-mind, that's another instance.

client/controllers/maps/eventMap.coffee
jgoley: True, didn't think about that. I'll revert back.

methods/incidentReports.coffee
jgoley: Question: do you think we should keep `addIncidentReportToEvent` in the "incidentReports" file or should it be with userEvent methods?

jgoley: I think we do need to pass the userEventId...I just forgot to when the methods are called. Mostly because we no longer store the `userEventId` on the incidents. So without the parameter, we'd have to query for the event's ID in order to update the event metadata?

jgoley:
Another question: Should we leave incidents on events when incidents are deleted?

methods/incidentReports.coffee
nathanathan: I think either file is fine.

Querying for the associated eventIds would be better that passing in the id. Since the incident may be shared by multiple events, events other than just the passed in event could need to have their meta-data updated.

nathanathan:
> Another question: Should we leave incidents on events when incidents are deleted?

I think we should remove them from the events. Maybe we should not allow them to be deleted until they are disassociated from their events. @Broham @toph-allen do you have a position on this?

server/updaters.coffee
```diff
@@ -10,14 +10,29 @@ feedSchema = require '/imports/schemas/feed'
...
+DATA_VERSION = 3
 AppMetadata = new Meteor.Collection('appMetadata')
 priorDataVersion = AppMetadata.findOne(property: "dataVersion")?.value
 
 Meteor.startup ->
   if priorDataVersion and priorDataVersion >= DATA_VERSION
     return
   console.log "Running database update code..."
+
+  Incidents.find(userEventId: $exists: true).forEach (incident) ->
```
nathanathan: This should happen at the end of the file so it doesn't break subsequent steps.

nathanathan:
When I try to edit incidents I'm getting these server errors:
```
I20170517-11:42:33.359(-4)? Exception while invoking method 'editIncidentReport' TypeError: Cannot read property '_id' of undefined
I20170517-11:42:33.359(-4)?     at [object Object].Meteor.methods.editUserEventLastIncidentDate (methods/userEvents.coffee:49:25)
I20170517-11:42:33.360(-4)?     at maybeAuditArgumentChecks (packages/ddp-server/livedata_server.js:1711:12)
I20170517-11:42:33.360(-4)?     at packages/ddp-server/livedata_server.js:1624:18
I20170517-11:42:33.360(-4)?     at [object Object]._.extend.withValue (packages/meteor/dynamics_nodejs.js:56:1)
I20170517-11:42:33.361(-4)?     at Server.apply (packages/ddp-server/livedata_server.js:1623:45)
I20170517-11:42:33.361(-4)?     at Server.call (packages/ddp-server/livedata_server.js:1566:17)
I20170517-11:42:33.362(-4)?     at [object Object].Meteor.methods.editIncidentReport (methods/incidentReports.coffee:47:14)
I20170517-11:42:33.362(-4)?     at maybeAuditArgumentChecks (packages/ddp-server/livedata_server.js:1711:12)
I20170517-11:42:33.363(-4)?     at packages/ddp-server/livedata_server.js:711:19
I20170517-11:42:33.363(-4)?     at [object Object]._.extend.withValue (packages/meteor/dynamics_nodejs.js:56:1)
I20170517-11:42:45.499(-4)? Exception while invoking method 'editIncidentReport' TypeError: Cannot read property '_id' of undefined
I20170517-11:42:45.501(-4)?     at [object Object].Meteor.methods.editUserEventLastIncidentDate (methods/userEvents.coffee:49:25)
I20170517-11:42:45.502(-4)?     at maybeAuditArgumentChecks (packages/ddp-server/livedata_server.js:1711:12)
I20170517-11:42:45.503(-4)?     at packages/ddp-server/livedata_server.js:1624:18
I20170517-11:42:45.505(-4)?     at [object Object]._.extend.withValue (packages/meteor/dynamics_nodejs.js:56:1)
I20170517-11:42:45.506(-4)?     at Server.apply (packages/ddp-server/livedata_server.js:1623:45)
I20170517-11:42:45.507(-4)?     at Server.call (packages/ddp-server/livedata_server.js:1566:17)
I20170517-11:42:45.508(-4)?     at [object Object].Meteor.methods.editIncidentReport (methods/incidentReports.coffee:47:14)
I20170517-11:42:45.509(-4)?     at maybeAuditArgumentChecks (packages/ddp-server/livedata_server.js:1711:12)
I20170517-11:42:45.510(-4)?     at packages/ddp-server/livedata_server.js:711:19
I20170517-11:42:45.514(-4)?     at [object Object]._.extend.withValue (packages/meteor/dynamics_nodejs.js:56:1)
```

userEventId should be removed from the incident schema.

In the Event Documents list incidents are not appearing when a document is selected.

The "add incident to event" button on the inbox doesn't seem to work if an incident is already added to an event.

jgoley:
I've fixed the issues you mentioned.
I also noticed rejected incidents were being displayed in the event pages so I updated the publications to only find accepted incidents.

toph-allen:
tl;dr I started writing a short comment and then wrote a long one. This contains some thoughts I've been mulling over and discussing about the Inbox/Documents UI cleanup.

---

I think I covered Jonathan's question when he and I were talking today. Incidents which are rejected/deleted should be removed from all events, because they are being marked as incorrect expressions of the data in the document.

In the discussion above, talking about Incident deletion, is "deletion" in this case what is referred to by the UI as "rejection"?

If so, this — plus my chats with @nathanathan and @jgoley yesterday and today about hiding Rejected Incidents in the table — make me think that we should probably just present this to the user as Deletion.

So, to recap, the changes I'm thinking about are (1) hiding Rejected Incidents from the table, and making them much less prominent in the document, and (2) changing the language we use around this to Deletion.

I say this because:
1. This is what we do to indicate that they are incorrect representations of the information in the document;
2. Thus, if they are Rejected/Deleted, they should be removed from all Events (if an Incident is part of an Event, we might want an alert dialogue saying something like, "This/These Incident(s) are part of the following Event(s). If you delete it/them, it/they will be removed from the Events."
3. "Delete" is a much clearer term, with existing connotations, than "Reject".

This switches the conceptual model we present to users:

- Currently, it is: `These Incidents represent the complete set of data points that have been created in the process of selecting what might or might not be the true case/death info in the document. Accepted ones have been either judged by our algorithm or a human to be true, and Rejected ones have been judged by our algorithm or a human to be false`. This kind of begs the question, "What *is* a Rejected Incident?" and "Why are all these random numbers assigned dates and locations?"
- With these changes, it would be "This table shows all the disease incident data in the document that can be structured; add or delete things as needed to make it properly reflect that."

We'd still keep Deleted Incidents around, because when a user clicks on an underlined number, or highlights a span containing one and hits "Add Incident", we're basically just promoting a "Deleted" incident to "Not Deleted".

jgoley:
I started working on incident "removal" before @toph-allen's comment.
34e8bd1 Removes or disassociates an incident from the event only. It's still in the db and more importantly an accepted incident of its parent document. I suppose in the future we can implement further changes/revisions?

nathanathan:
It looks like disassociation feature you're working on does not conflict with the deletion feature @toph-allen described. We should be careful to use the term "disassociate" for it to avoid confusion.

Hiding the rejected incidents should be doing separately from this PR. There is already a lot going on in this one. I created a pivotal task for it: https://www.pivotaltracker.com/story/show/145665197

jgoley:
@nathanathan I've noticed sorting events by incident count is a little broken.
I tried passing in a `sortFn` method but it didn't seem to help.
Any ideas?

client/controllers/events/curatedEvent.coffee
```diff
@@ -1,15 +1,48 @@
...
+      document.title = "Eidr-Connect: #{userEvent.eventName}"
+      userEvent.incidents.forEach (incident) =>
+        @incidentIds.upsert id: incident.id,
+          id: incident.id
+
+  @autorun =>
+    incidentIds = _.pluck(@incidentIds.find().fetch(), 'id')
+    @subscribe "eventArticles", userEventId, incidentIds
+    if incidentIds
+      @subscribe 'eventIncidents', incidentIds, =>
```
nathanathan: Putting these subscriptions in an autorun adds a round trip and slows down the load times. They could both take only the userEventId as a parameter and use that to figure out which incidents to return.

nathanathan:
I think sorting isn't working since the count is a computed property and can't be done by the mongo query that is used for pagination.

It looks like filtering out unaccepted incidents is causing many of the older incidents with no "accepted" property to be removed. I don't think this filter is necessary for curated events. Rejecting an incident should disassociate it from all events.

Also, could you remove the articleCount property from the schema.

jgoley:
Should we disable sorting on Incident Counts then? Looks sorting on server-side collections is on the value of the key. Or can we work around this?

nathanathan:
@jgoley Disabling sorting on the incident count column is okay.

jgoley:
Are the tests failing because they are running before the db update?

jgoley:
Oh its because of the call of `addIncidentReport` in the fixtures? I'll fix that.

jgoley:
@nathanathan Looks like tests started failing because the doc processing time was a little longer than the wait time. I increased it by 3 seconds. Seems like there should be a better way to fix this but maybe not worth the time?

nathanathan:
I think the wait for time limit can be set globally in the chimp file here: https://github.com/ecohealthalliance/eidr-connect/blob/master/.config/chimp.js#L61

jgoley:
Ohh—I thought the global wait time was shorter.  I removed the time argument in the function causing the issue in favor of the default.

jgoley:
@nathanathan 1cd5dbd addresses your comment about the subscriptions in curatedEvent.
There is an issue with reactivity which was my inspiration for tracking the ids in a local collection. Incident removal seems to be reactive so I'm not sure what's going on. Any ideas?

nathanathan:
The publications will not reactively update for all the collections queried, but only for the collection published. Let's try using [publish-composite](https://atmospherejs.com/reywood/publish-composite) for this. The meteor guide recommends using it to publish relational data: https://guide.meteor.com/data-loading.html#publishing-relations

jgoley:
@nathanathan Good call with using publish-composite!
3848568 takes care of the necessary updates.
I created two new collections (`EventIncidents` and `EventArticles`) to reduce any potential interference from when we subscribe to article incidents (when we present article details).

nathanathan:
The event map doesn't seem to be loading any markers.

jgoley:
@nathanathan Strange—I'll take a look and add a test to ensure markers are placed on the map.

jgoley:
@nathanathan I'm working on a fix for document issue. Due to the implied association changes, users can't really remove documents of the event has incidents from the article. I'm thinking we only allow users to remove if no incidents? Only the article was associated?

nathanathan:
Only allowing documents with no incidents associated with the event to be removed sounds good to me.

nathanathan:
👍 Looks good!

PR #365, opened 2017-05-18
Remove existing incident type props when incident types changes

jgoley:
[Fixes this bug](https://www.pivotaltracker.com/story/show/145504331)
Looks like there are 19 incidents that have both `cases` and `deaths` properties but most, if not all, instances have the same value. This will cause issues if the incidents are edited in the future. We may want to review these incidents and manually change the type?

### DIFF
--- a/client/controllers/incidentReports/incidentModal.coffee
+++ b/client/controllers/incidentReports/incidentModal.coffee
@@ -83,20 +83,8 @@ Template.incidentModal.events
         instance.submitting.set(false)
 
     if @edit
-      _incident = _.extend({}, @incident, incident)
-
-      # Remove existing type props if user changes incident type
-      fieldsToRemove = []
-      if incident.cases
-        fieldsToRemove = ['deaths', 'specify']
-      else if incident.deaths
-        fieldsToRemove = ['cases', 'specify']
-      else if incident.specify
-        fieldsToRemove = ['cases', 'deaths']
-      fieldsToRemove.forEach (field) ->
-        delete _incident[field]
-
-      Meteor.call 'updateIncidentReport', _incident, fieldsToRemove, (error, result) ->
+      incident._id = @incident._id
+      Meteor.call 'editIncidentReport', incident, (error, result) ->
         if not error
           $('.reactive-table tr').removeClass('open')
           $('.reactive-table tr.details').remove()
--- a/client/controllers/incidentReports/incidentModal.coffee
+++ b/client/controllers/incidentReports/incidentModal.coffee
@@ -83,8 +83,20 @@ Template.incidentModal.events
         instance.submitting.set(false)
 
     if @edit
-      incident = _.extend({}, @incident, incident)
-      Meteor.call 'editIncidentReport', incident, (error, result) ->
+      _incident = _.extend({}, @incident, incident)
+
+      # Remove existing type props if user changes incident type
+      fieldsToRemove = []
+      if incident.cases
+        fieldsToRemove = ['deaths', 'specify']
+      else if incident.deaths
+        fieldsToRemove = ['cases', 'specify']
+      else if incident.specify
+        fieldsToRemove = ['cases', 'deaths']
+      fieldsToRemove.forEach (field) ->
+        delete _incident[field]
+
+      Meteor.call 'updateIncidentReport', _incident, fieldsToRemove, (error, result) ->
         if not error
           $('.reactive-table tr').removeClass('open')
           $('.reactive-table tr.details').remove()
--- a/client/controllers/incidentReports/suggestedIncidentModal.coffee
+++ b/client/controllers/incidentReports/suggestedIncidentModal.coffee
@@ -80,10 +80,9 @@ Template.suggestedIncidentModal.events
         $set:
           accepted: false
     else
-      Meteor.call 'updateIncidentReport', {
-        _id: instance.incident._id
-        accepted: false
-      }, (error, result) =>
+      incident = isntance.incident
+      incident.accepted = false
+      Meteor.call 'editIncidentReport', instance, (error, result) =>
         if error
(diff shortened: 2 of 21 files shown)


client/controllers/incidentReports/incidentModal.coffee
```diff
@@ -78,8 +78,20 @@ Template.incidentModal.events
           notify('error', errorString)
 
     if @edit
-      incident = _.extend({}, @incident, incident)
-      Meteor.call 'editIncidentReport', incident, (error, result) ->
+      _incident = _.extend({}, @incident, incident)
+
+      # Remove existing type props if user changes incident type
+      fieldsToRemove = []
+      if incident.cases and not @incident.cases
```
nathanathan: Why are the "and not" clauses needed?

jgoley: Oh, good call—excessive.

methods/incidentReports.coffee
```diff
@@ -21,15 +21,22 @@ Meteor.methods
...
     if not Roles.userIsInRole(user._id, ['admin'])
       throw new Meteor.Error("auth", "User does not have permission to edit incidents")
     incident.modifiedByUserId = user._id
     incident.modifiedByUserName = user.profile.name
-    res = Incidents.update({_id: _id}, {$set: incident})
+    updateOperators =
+      $set: incident
+    if fieldsToRemove.length
+      fieldsToUnset = {}
+      fieldsToRemove.forEach (field) ->
```
nathanathan: We should limit the removable fields to only cases, deaths and specify so things like modification dates and users can't be altered. Determining the fields to remove in the method might work better than passing them in since we won't need to worry about bad data and it the field removal logic won't need to be repeated in other places where the incident type is altered.

jgoley: The incident obj we pass to the method is a combination of the old and new (user entered via the form). If the type is changed it will have both properties. So would you recommend querying for the old and doing the a comparison in the method?

nathanathan: I don't see the reason for querying for the old incident. If one of the cases/deaths/specify fields is passed in then we can unset the other two.

jgoley: I might be looking at this the wrong way but if we combine the old and new incidents ([here](https://github.com/ecohealthalliance/eidr-connect/blob/master/client/controllers/incidentReports/incidentModal.coffee#L81)) then in the method, we won't know which is new and which should be removed?

nathanathan: Ah, I see the issue. However, when using updateIncidentReport instead of edit I don't think there is a reason to extend the old incident report.

jgoley: Good call. I wonder though if there are any properties we'd loose? I'll check and go that route.

jgoley: Looks like using `updateIncidentReport` adds the new incident type property and leaves the existing. `editIncidentReport` will replace the entire object which is partially what we want to do but we loose several properties in the process (addedByUserName, addedByUserId, addedDate, etc.) Should we go this route and add those properties to the form-based incident object we pass to the method?

nathanathan: I think either method would be fine. However, editIncidentReport ought to prevent users from submitting properties that are maintained by the application like added_____. Those properties should not be passed from the client. The editIncidentReport method should look up the prior incident report and add the application properties to the submission in order to preserve them.

nathanathan:
I'm not able to edit any of the fields besides the incident type. Could you add an incident editing test?

I'm also seeing this console error:
```
Exception in template helper: TypeError: Cannot read property 'case' of undefined
    at getIncidentSnippet (http://nathan.eha.io/app/app.js?hash=905be3e881e855024698fbd6325bc9d7ec7a44d2:15703:37)
    at Object.incidentSnippet (http://nathan.eha.io/app/app.js?hash=905be3e881e855024698fbd6325bc9d7ec7a44d2:7465:39)
```

jgoley:
Yep, I'll add a test of incident editing and fix that issue.
I think the console error might be fixed in #363 [here](https://github.com/ecohealthalliance/eidr-connect/pull/363/files#diff-101292d00ff3013007d2e031b238a216R4).

methods/incidentReports.coffee
```diff
@@ -40,9 +41,26 @@ Meteor.methods
...
+    # Remove existing type props if user changes incident type and merge incident
+    # from client with existing incident
+    if incident.cases
+      fieldsToRemove = deaths: true, specify: true
+    else if incident.deaths
+      fieldsToRemove = cases: true, specify: true
+    else if incident.specify
+      fieldsToRemove = cases: true, deaths: true
+    existingIncident = Incidents.findOne(incident._id)
+    updatedIncident = _.extend(existingIncident, incident)
```
nathanathan: Keeping the properties from the old incident makes this method almost the same as updateIncidentReport. We should combine them into a single method.

nathanathan:
I can't edit incidents in the inbox when I try to change their type from cases to deaths.

client/controllers/incidentReports/suggestedIncidentModal.coffee
```diff
@@ -80,10 +80,9 @@ Template.suggestedIncidentModal.events
         $set:
           accepted: false
     else
-      Meteor.call 'updateIncidentReport', {
-        _id: instance.incident._id
-        accepted: false
-      }, (error, result) =>
+      incident = isntance.incident
```
nathanathan: There is a typo in instance.

jgoley:
@nathanathan I fixed the typo and removed the typo and fixed the issue with saving incidents in the inbox—the incident's id was undefined.

nathanathan:
When the incident status is deselected the change isn't saved.

nathanathan:
👍 Looks good

PR #366, opened 2017-05-18
Incident resolution and case/death rate plot

nathanathan:
This adds an incident report resolution algorithm and a plot that shows the resolved cases/death rates for the top level locations in the incident report collection.


jgoley:
This is awesome!
I'm curious why you use `console.assert` here and there? 
I'm going to make some minor styling updates. I suppose the layout will change a fair amount when we revise the event page so I won't make too many changes.

nathanathan:
Thanks! I used the console.asserts to track down a few bugs during the course of development. I left them in since their console errors could be helpful in identifying the source of future problems.

jgoley:
Sounds good. I forgot to mention earlier that there are a few typos in the comments. ([here](https://github.com/ecohealthalliance/eidr-connect/pull/366/files#diff-fda1ccf366a1dfe791d65097511679dfR27)) Not a big deal at all but thought I'd mention it.

jgoley:
I made some small changes.
Ideally I'd like to use colors in the chart closer to what we've established in the app but considering there could be a bunch of locations, this may not be feasible.

nathanathan:
@jgoley I replaced highcharts with rickshaw because of highchart's license restrictions. This required making some changes to the plot. You might want to take another look at it.

jgoley:
I've updated the styles of the chart—mostly updating the legend.
I noticed the Rickshaw chart isn't responsive? Looks like its not supported: https://github.com/shutterstock/rickshaw/issues/483
This may be a problem.

jgoley:
Also, on events with many incidents, the steps in the color range can be indistinguishable.
And when switching from Death Rate back to Case Rate, there is a relatively long delay with no indication to the user that something is loading or processing.
This event is a good example: http://localhost:3000/events/curated-events/gtT3A6ozJsvzHdzWp

nathanathan:
@jgoley I added a loading indicator. We might be able to eventually do something about the lack of distinct colors for all the countries in this case by adding data for continent level containment.

PR #367, opened 2017-05-22
Hide rejected incidents table

nathanathan:
Pivotal story: https://www.pivotaltracker.com/story/show/145665197

This also fixes a bug where the suggestedIncident modal's reject button wasn't working.

### DIFF
--- a/client/controllers/incidentReports/annotatedContent.coffee
+++ b/client/controllers/incidentReports/annotatedContent.coffee
@@ -1,4 +1,7 @@
-import { annotateContentWithIncidents } from '/imports/ui/annotation'
+import {
+  annotateContentWithIncidents,
+  buildAnnotatedIncidentSnippet } from '/imports/ui/annotation'
+import Incidents from '/imports/collections/incidentReports.coffee'
 
 POPUP_DELAY = 100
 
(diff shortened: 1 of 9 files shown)


client/controllers/incidentReports/annotatedContent.coffee
```diff
@@ -27,7 +30,11 @@ Template.annotatedContent.onDestroyed ->
 Template.annotatedContent.helpers
   annotatedContent: ->
     selectedAnnotationId = Template.instance().data.selectedAnnotationId.get()
-    annotateContentWithIncidents(@content, @incidents.fetch(), selectedAnnotationId)
+    annotateContentWithIncidents(@content, @incidents.fetch().map((incident)->
+      if not incident.accepted
+        incident.uncertainCountType = true
```
Broham: Is `uncertainCountType` necessary? It seems like we could just be checking whether an incident is accepted everywhere we are using it instead?  If we did go this route we would need to refactor in a few other places.

nathanathan: On the extract-incidents page uncertain incidents (where grits didn't think it was an incident) and rejected incidents (where the user removed an incident grits approved of) still have different highlights. The variable is necessary to maintain that functionality.

PR #368, opened 2017-05-23
Fix Create Event and Incident modal issues

jgoley:
Fixes [this bug](https://www.pivotaltracker.com/story/show/145689715)
Also changes how we pass in articles to the incident modal/form. Before, if the incident existed, we passed in an array of articles. I've changed this to passing in an incident object with an articleId (if creating a new incident) and checking for its existence.

### DIFF
--- a/client/controllers/curatorInbox/addToEvent.coffee
+++ b/client/controllers/curatorInbox/addToEvent.coffee
@@ -1,6 +1,7 @@
 UserEvents = require '/imports/collections/userEvents.coffee'
 Articles = require '/imports/collections/articles.coffee'
 { notify } = require '/imports/ui/notification'
+{ pluralize } = require '/imports/ui/helpers'
 
 Template.addToEvent.onCreated ->
   @subscribe('userEvents')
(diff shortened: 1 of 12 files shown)


jgoley:
@nathanathan I think 2753b0a and db3011e fix the other issues mentioned in the bug.

client/controllers/curatorInbox/addToEvent.coffee
```diff
@@ -24,15 +25,19 @@ Template.addToEvent.onRendered ->
       placeholder: 'Search for an Event...'
       minimumInputLength: 0
 
-    $(document).on 'click', '.add-new-event', (event) ->
+    $(document).on 'click', ".add-new-event-#{instanceData.objNameToAssociate}", (event) ->
```
nathanathan: Do you know why a document click handler is being used here instead of a template click handler? I think this might cause bugs since nothing appears to clean up the old handlers.

jgoley: Seems like the issue was related to the select2 related DOM elements being detached from Blaze or the template. I tried using a template click handler but the event never fires.

nathanathan:
👍

PR #369, opened 2017-05-30
Add associated events column to inbox

nathanathan:
Story: https://www.pivotaltracker.com/story/show/145689277

This also fixes the modal lingering bug:
https://www.pivotaltracker.com/story/show/144528655


jgoley:
Do you think its important for the user to see the number of event associations in the table? I'm thinking this may be a good time to create a secondary content/action area so that we don't clutter the table? (Like in the incident table on the event page. We could recycle the associated logic and such)

nathanathan:
@jgoley I do think showing the number of associations is important. Without it associating incidents with multiple events is difficult.

jgoley:
@nathanathan I think this area of the UI needs a good bit of work so I'll create a chore to fix it up.

PR #370, opened 2017-05-30
Click incident in doc to reveal actions

jgoley:
[PT Story](https://www.pivotaltracker.com/story/show/145857845)

I intended to write tests but am currently having trouble running them locally.I encountered an issue with `fibers` which I've seen before but forgot how to resolve. 
```/Users/Goley/code/eidr-connect/node_modules/chimp/node_modules/cucumber/node_modules/fibers/bin/darwin-x64-47/fibers.node` is missing.```
I'll try to fix in the morning or if this is okay we could break tests out into a separate chore.

### DIFF
--- a/client/controllers/incidentReports/annotatedContent.coffee
+++ b/client/controllers/incidentReports/annotatedContent.coffee
@@ -46,13 +46,26 @@ Template.annotatedContent.events
         scrolled: instance.scrolled
         selecting: instance.selecting
         popupDelay: POPUP_DELAY
+        view: 'newIncidentFromSelection'
       Blaze.renderWithData(
-        Template.newIncidentFromSelection,
+        Template.popup,
         data,
         $("#{instance.data.relatedElements.parent}")[0]
         $("#{instance.data.relatedElements.sibling}")[0]
       )
     else
+      data =
+        source: instance.data.source
+        scrolled: instance.scrolled
+        selecting: instance.selecting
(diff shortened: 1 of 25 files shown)

### CODE REVIEW

File: client/controllers/incidentReports/annotatedContent.coffee [hunk 1]
```diff
@@ -64,6 +62,22 @@ Template.annotatedContent.events
         # Set reactive variable that's handed down from curatorSourceDetails and
         # shared with the incidentTable templates to the clicked annotation's ID
         instance.data.selectedAnnotationId.set(annotationId)
+        data =
+          incidentId: annotationId
+          source: instance.data.source
+          scrolled: instance.scrolled
+          showPopup: instance.showPopup
+          popupDelay: POPUP_DELAY
```
nathanathan: I think it would make more sense to move POPUP_DELAY to the popup template than pass it as a parameter.

client/controllers/incidentReports/popup.coffee
```diff
@@ -0,0 +1,61 @@
+import { buildAnnotatedIncidentSnippet } from '/imports/ui/annotation'
```
nathanathan: Looks like these imports are unused.

jgoley:
I just discovered a bug with selecting the incident in the little popup. Will fix asap.

nathanathan:
👍 Looks good

PR #371, opened 2017-05-31
Only require document url or content

nathanathan:
Story: https://www.pivotaltracker.com/story/show/145932369

This also updates meteor and refactors the template helpers file.


jgoley:
I updated a few things in da27832
I'm seeing an error when saving an edited source: 
```
Uncaught TypeError: Cannot read property 'value' of undefined
    at Object.click .save-source (sourceModal.coffee:181)
```
Also: If the user does not enter a title, there will be a several wholes in the UI. I suppose this is okay? should we show "Untitled Document" or something similar?

nathanathan:
I'm using the url as a replacement when the title isn't available. I think my last commit resolves the source editing bug you found.

jgoley:
What if the document is user entered text without a URL? 

Looks like the url only is shown in the table, not in the details. 
Also looks like if a date is not defined by the user, the publish date is shown as "invalid Date" in the document details:
screen shot 2017-06-01 at 11 30 47 am

nathanathan:
@jgoley I fixed the missing details you found and handled documents without a title or url by showing their first 30 characters as a title.

PR #372, opened 2017-06-01
Ui enhancement fixed incident table header

jgoley:
This was a little more complex because I created separate template for the logic and such.

I branched from #370 to take advantage of some  overlapping changes it introduces.
[PT Story](https://www.pivotaltracker.com/story/show/146163505)

### DIFF
--- a/client/views/curatorInbox/sourceIncidentReportsHeader.jade
+++ b/client/views/curatorInbox/sourceIncidentReportsHeader.jade
@@ -1,0 +1,19 @@
+template(name="sourceIncidentReportsHeader")
+  .curator-source-details--sub-header
+    button.btn.btn-sm.btn-danger.delete(
+      disabled="{{#unless incidentsSelectedCount}} true {{/unless}}") Delete
+    button.btn.btn-sm.btn-primary.show-addEvent(
+      disabled="{{#unless incidentsSelectedCount}} true {{/unless}}") Add to Event
+    button.btn.btn-sm.btn-default.select.on-right(
+      disabled="{{#unless incidents.count}} true {{/unless}}"
+      class="{{#if allSelected}} deselect-all {{else}} select-all {{/if}}")
+      if allSelected
+        | Deselect All
+      else
+        | Select All
+  if addEvent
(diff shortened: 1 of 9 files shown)


client/views/curatorInbox/sourceIncidentReportsHeader.jade
```diff
@@ -0,0 +1,22 @@
+template(name="sourceIncidentReportsHeader")
+  .curator-source-details--sub-header
+    button.btn.btn-sm.btn-danger.delete(
+      disabled="{{#unless incidentsSelectedCount}} true {{/unless}}")
+      if deleting
```
nathanathan: I don't think we need a deleting loader. We can use latency compensation to prevent delays.

packages/rickshaw/.npm/package/.gitignore
```diff
@@ -0,0 +1 @@
+node_modules
```
jgoley: hm, I don't recall adding that. I'll remove it.

client/views/curatorInbox/sourceIncidentReportsHeader.jade
jgoley: Can we take advantage of the latency compensation here since the UI changes depend on an update to the local collection. Should we remove the incidents from the local collection before the method call assuming there will be no error?

jgoley: Oh, nevermind...the local collection only controls what's selected.
I added it because there's a significant delay locally for me.

jgoley: I went ahead and removed it, If its a problem I suppose we can fix it.

packages/rickshaw/.npm/package/.gitignore
jgoley: Every time I run the application, meteor adds the `.npm` folder. I'll just add it to my global gitignore if its not an issue on your end.

nathanathan:
👍 Looks good

PR #373, opened 2017-06-02
Fix inbox search box position

jgoley:
[PT Story](https://www.pivotaltracker.com/story/show/144115899)

I couldn't replicate the bug mentioned in the story. It may be an IE thing? @toph-allen Do you still have windows setup on your computer? If so, could you try to replicate this?


toph-allen:
I'll try on IE.

jgoley:
Due to the issues with the transitions/animations, I'll switch to hide/close.

nathanathan:
👍 Looks good

PR #374, opened 2017-06-05
Disassociate incidents on deletion

nathanathan:
Story: https://www.pivotaltracker.com/story/show/146455441

When the incidents to be deleted have associated been with events the user will be prompted to confirm the deletion.

This also fixes this grits error handling bug: https://www.pivotaltracker.com/story/show/146450531


client/views/curatorInbox/curatorSourceDetails.jade
```diff
@@ -53,9 +53,20 @@ template(name="curatorSourceDetails")
               source=source)
           if notifying
             .veil
+        else if enhancements.error
+          .curator-inbox--warning
+            i.fa.fa-warning
+            h2=enhancements.error
+            .btn.retry Retry
+        else if enhancements.processingStartedAt
```
jgoley: Is there a reason we show the notification when the document is still processing rather than continuing to show the animated loading indicator?

jgoley:
Did you remove the incident removal method from the deleteConfirmation because of dealing with selectedIncidents? I like handing down the `onConfirm` method. Maybe we could move the deleteConfirmationModal down this path. Seems like we could avoid creating another template and use the deleteConfirmationModal for this?

client/views/curatorInbox/curatorSourceDetails.jade
nathanathan: Yes, sometimes the document will get stuck in the processing state and this makes it possible to retry it.

nathanathan:
@jgoley I combined the deleteConfirmationModal with the confirmationModal. I removed the incident removal method from deleteConfirmation template because the onConfirm callback is used instead.

client/controllers/articles.coffee
```diff
@@ -111,12 +111,14 @@ Template.articles.events
...
-    Modal.show 'deleteConfirmationModal',
-      userEventId: instance.data.userEvent._id
-      objNameToDelete: 'source'
-      objId: sourceId
-      displayName: EventArticles.findOne(sourceId).title
-    instance.$(event.currentTarget).tooltip('destroy')
+    Modal.show 'confirmationModal',
+      html: Spacebars.SafeString(Blaze.toHTMLWithData(
+        Template.deleteConfirmationModalBody,
+        objNameToDelete: 'Document'
```
jgoley: Looks like we need to give the template the doc title as `displayName`?

PR #375, opened 2017-06-05
Fix adding manual annotation - keep annotation on save

jgoley:
[PT Bug](https://www.pivotaltracker.com/story/show/146605755)
- Refactor `newIncidentFromSelection` controller logic
- Use simpleSchema `clean` method to filter out props not in schema

### DIFF
--- a/client/controllers/incidentReports/suggestedIncidentModal.coffee
+++ b/client/controllers/incidentReports/suggestedIncidentModal.coffee
@@ -4,10 +4,9 @@ incidentReportSchema = require '/imports/schemas/incidentReport.coffee'
 { stageModals } = require '/imports/ui/modals'
 
 Template.suggestedIncidentModal.onRendered ->
-  instance = @
-  Meteor.defer ->
+  Meteor.defer =>
     # Add max-height to snippet if it is taller than form
-    formHeight = instance.$('.add-incident--wrapper').height()
+    formHeight = @$('.add-incident--wrapper').height()
     $snippet = $('.snippet--text')
     if $snippet.height() > formHeight
       $snippet.css('max-height', formHeight)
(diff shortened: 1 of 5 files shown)


client/controllers/incidentReports/suggestedIncidentModal.coffee
```diff
@@ -4,10 +4,9 @@ incidentReportSchema = require '/imports/schemas/incidentReport.coffee'
 { stageModals } = require '/imports/ui/modals'
 
 Template.suggestedIncidentModal.onRendered ->
-  instance = @
```
nathanathan: I see some remaining instance references.

methods/incidentReports.coffee
```diff
@@ -13,6 +13,7 @@ Meteor.methods
   addIncidentReport: (incident, userEventId) ->
     user = Meteor.user()
     checkPermission(user._id)
+    incident = incidentReportSchema.clean(incident)
```
nathanathan: I don't think we should clean the incident here. If extra parameters are being added it would be better for it to cause a validation error so we can catch the error. Cleaning incidents before passing them to this method would be fine.

client/controllers/incidentReports/suggestedIncidentModal.coffee
jgoley: In `onRendered`? I think there was only one?
I did found [this](https://github.com/ecohealthalliance/eidr-connect/blob/2343ba7a6bd8721b54f551c87e06c11f77d6d2f7/client/controllers/incidentReports/suggestedIncidentModal.coffee#L29) which should be `incident._id`

nathanathan: My mistake, this looks good.

nathanathan:
👍

PR #376, opened 2017-06-05
Fix unaccepted incidents showing in doc details 146606131

jgoley:
[PT Bug](https://www.pivotaltracker.com/story/show/146606131)
Also adds counts for document locations and incidents in document details and incident count in inbox document details.

I also updated the wording of a few helpers from source to document to I wonder if we should start renaming references to documents in the code? We refer to them as _documents_, _articles_ and _sources_. 😕

### DIFF
--- a/client/controllers/articles.coffee
+++ b/client/controllers/articles.coffee
@@ -67,7 +67,9 @@ Template.articles.helpers
     source.title or source.url or (source.content?.slice(0,30) + "...")
 
   incidentsForSource: (source) ->
-    Incidents.find(articleId: Template.instance().selectedSourceId.get())
+    Incidents.find
+      articleId: Template.instance().selectedSourceId.get()
+      accepted: true
 
   locationsForSource: (source) ->
     locations = {}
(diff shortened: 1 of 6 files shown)


server/publications.coffee
```diff
@@ -37,6 +37,7 @@ Meteor.publish 'articleIncidents', (articleId) ->
   Incidents.find
     articleId: articleId
     deleted: $in: [null, false]
+    accepted: true
```
nathanathan: The unaccepted incidents are used on the curator inbox to underline potential counts.

jgoley: Oh yeah, my bad. I'll update the query that feeds the table.

jgoley: Or only filter out deleted in the publication, rather.

nathanathan:
The incidents aren't showing at all now.

jgoley:
@nathanathan Sorry about that, I suppose it was a merge error.

nathanathan:
👍

PR #381, opened 2017-06-06
Fix widths, margins and padding of suggestedIncidentModal

jgoley:
Fixes some visual bugs I discovered after the updates to suggestedIncidentModal.


nathanathan:
👍

PR #382, opened 2017-06-07
Prevent user emails from being published

nathanathan:
This prevents user emails from being published to the world. Emails are still visible to admins on the account management page.


jgoley:
This is a much better approach! 👍

PR #383, opened 2017-06-08
Fix top level locations not combining rates from all sub-locations

nathanathan:
In the resolved incident plot the top locations previously used the largest rate from the sub-location if no rate was defined for the top location. This lead to rates being lower than they should be because the rates from disjoint sub-locations should have been combined. This PR fixes the issue by ensuring top location rates are defined in every sub-interval.

This also improves the interval detail view by displaying sub-components hierarchically.


jgoley:
👍 
I made some small changes to the UI.
I'm curious how we run the test you added for the location tree.

nathanathan:
@jgoley To run the unit tests launch meteor with the command `meteor test --driver-package practicalmeteor:mocha` then visit the webpage.

PR #385, opened 2017-06-15
Add modal for creating new geonames

nathanathan:
Pivotal link: https://www.pivotaltracker.com/story/show/144206547


client/controllers/addGeonameModal.coffee
```diff
@@ -0,0 +1,42 @@
...
+      featureCode: 'PPL'
+      latitude: latitude
+      longitude: longitude
+    Meteor.call('addGeoname', geoname, (error)->
+      if error
+        notify('error', 'There was a problem creating the geoname.')
+        return
+      instance.onAdded(geoname)
+      Modal.hide(instance)
+    )
```
jgoley: Maybe we should remove the enclosing parens to stay stylistically consistent?

client/controllers/incidentReports/incidentForm.coffee
```diff
@@ -1,7 +1,11 @@
...
-validator = require 'bootstrap-validator'
-{ keyboardSelect, removeSuggestedProperties, diseaseOptionsFn } = require '/imports/utils'
+import Articles from '/imports/collections/articles.coffee'
+import createInlineDateRangePicker from '/imports/ui/inlineDateRangePicker.coffee'
+import validator from 'bootstrap-validator'
+import {
+  keyboardSelect,
+  removeSuggestedProperties,
+  diseaseOptionsFn,
+  locationOptionsFn } from '/imports/utils'
```
jgoley: Looks like `locationOptionsFn` is imported but not used?

client/controllers/incidentReports/incidentModal.coffee
```diff
@@ -44,9 +44,10 @@ Template.incidentModal.events
     # Submit the form to trigger validation and to update the 'valid'
     # reactiveVar — its value is based on whether the form's hidden submit
     # button's default is prevented
-    instance.submitting.set(true)
     $('#add-incident').submit()
-    return unless instance.valid.get()
+    if instance.valid.get()
+      console.log "invalid"
```
jgoley: Do we want to leave this log?

packages/coordinate-selector/coordinateSelector.coffee
```diff
@@ -0,0 +1,60 @@
...
+    Template.instance().latLon.get()?[0]
+  longitude: ->
+    Template.instance().latLon.get()?[1]
+  required: ->
+    Template.instance().required
+
+Template.coordinateSelector.events
+  'click .leaflet-clear': (event, t) ->
+    t.latLon.set(null)
+  'change .lat, change .lon': (event, t) ->
```
jgoley: For consistency, I think we should change the reference to the template instance (`t`) to `instance`.
Same for first event.

jgoley:
I made some style and UX-ish updates.
I wasn't able to save the incident after adding a custom location. Maybe the form validator thinks the form is invalid?

nathanathan:
@jgoley I can't reproduce the incident saving bug you mentioned. Could you go into more detail about which controls you're using to create in the incident and the values you're using in all the form fields.

jgoley:
After I add a custom location (and make no other changes to the incident) the form submit is blocked. For some reason `isDeafaultPrevented()` is true so the `valid` reactiveVar is not set to true. If I remove the custom location the form will submit.
If I remove the existing location and add the same one back I get this error on submit: 
```
Uncaught TypeError: Cannot read property 'alternateNames' of undefined
    at Object.incidentReportFormToIncident (utils.coffee:96)
```
(This may be an unrelated bug.)

The form submit is blocked also if I add additional locations provided in the select2 list. So maybe its an issue related to submitting multiple locations? But if I add a custom location, then remove the existing location, the form still fails to submit.

I'm also seeing this in the console when the modal opens: `The select2('close') method was called on an element that is not using Select2.` Its thrown in an autorun in a try block. Unrelated but maybe we should check if the element is using select2 first or defer?

client/controllers/incidentReports/incidentModal.coffee
```diff
@@ -45,8 +45,8 @@ Template.incidentModal.events
     # reactiveVar — its value is based on whether the form's hidden submit
     # button's default is prevented
     $('#add-incident').submit()
-    if instance.valid.get()
-      console.log "invalid"
+    if not instance.valid.get()
+      console.error "Invalid incident form"
```
jgoley: Could we, should we use `console.assert` here. I suppose it's about the same?

jgoley:
I'm still having trouble submitting the form after adding a custom location. I can try to look into it if you'd like. It's weird that you aren't experiencing the issue.

I also noticed a UX issue. In testing adding a custom location, I often would click on one of the provided locations when trying to click on the button (because the list is rendered just as I'm clicking) If possible, maybe we should add a loading indicator. I can also try to add this if you'd like.

jgoley:
I decided to move away from the bootstrap validator in favor of parsleyjs because I think the issue was a bug somewhere with it. Instead of checking whether the default is prevented on the form, with parsley we can check the form validity by calling a method.
I added parsely as a package instead of importing the module because installing the npm upgrades jquery which causes issues with bootstrap. I suppose we could specify the version of package? Not sure which is better?
I'll add a chore to refactor the other forms using bootstrap validator and remove the package.

nathanathan:
@jgoley The changes you made are broke the location selector. There are also a bunch of `this.removeLoader is not a function` errors in the console. I would like to remove the loader code since it's an additional part that can break and the locations take a trivial amount of time to load.

jgoley:
@nathanathan You are right about the loading indicator—I removed the commit which added it.

PR #386, opened 2017-06-19
Ui enhancement update incident list

jgoley:
[PT Chore](https://www.pivotaltracker.com/story/show/146374497)
[Related PT Story](https://www.pivotaltracker.com/story/show/145854209)


client/controllers/incidentReports/incidentSecondaryDetails.coffee
```diff
@@ -0,0 +1,30 @@
...
+Template.incidentSecondaryDetails.helpers
+  detailsOpen: ->
+    Template.instance().detailsOpen.get()
+
+  firstLocationName: ->
+    firstLocation = Template.instance().data.locations[0]
+    firstLocation?.countryName or
+      firstLocation?.admin1Name or
+      firstLocation?.admin2Name or
+      firstLocation?.name
```
jgoley: @nathanathan Is there a better way to do this?

nathanathan: Every location has a name so you can always use that. Is there a reason for using the country name instead? I don't think the admin names will be defined if there is no country name, so you could leave those out.

nathanathan:
Looks good but there are a few bugs:
* The dates are incorrect.
* The associated events lost their links.
* I'm seeing a lot of `formatLocation is not defined` errors in the console.

nathanathan:
👍

PR #387, opened 2017-06-20
Add cumulative mode to resolved incident plot

nathanathan:
Pivotal Story: https://www.pivotaltracker.com/story/show/146605873

### DIFF
--- a/client/controllers/resolvedIncidentsPlot.coffee
+++ b/client/controllers/resolvedIncidentsPlot.coffee
@@ -192,22 +192,22 @@ Template.resolvedIncidentsPlot.helpers
     Template.instance().loading.get()
 
 Template.resolvedIncidentsPlot.events
-  "click .incident-type-selector .cases": (e, instance)->
+  "click .incident-type-selector .cases": (event, instance)->
     instance.incidentType.set("cases")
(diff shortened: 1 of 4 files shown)


jgoley:
Looks like there is a chance the graph can be empty. I wonder if we should disable the button or somehow indicate there isn't enough data to show the plot?

Looks like the buttons break weirdly at smaller screen widths/resolutions. I'll work on this in the morning.

client/controllers/resolvedIncidentsPlot.coffee
```diff
@@ -153,9 +194,19 @@ Template.resolvedIncidentsPlot.helpers
 Template.resolvedIncidentsPlot.events
   "click .incident-type-selector .cases": (e, instance)->
```
jgoley: We should probably use `event, instance` as the parameters to stay consistent?

jgoley:
@nathanathan It looks like the tests are failing because of a property on the incident which isn't  allowed by the schema. It appears to fail in master and in #386.

nathanathan:
@jgoley Thanks for looking into the test failure. I think it was caused by some properties I added to the geoname lookup api responses for #385

PR #388, opened 2017-06-21
Add loading spinner to CSV download button

jgoley:
[PT Bug](https://www.pivotaltracker.com/story/show/144726317)

### DIFF
--- a/client/controllers/downloadCSVModal.coffee
+++ b/client/controllers/downloadCSVModal.coffee
@@ -1,10 +1,26 @@
+Template.downloadCSVModal.onCreated ->
+  @prepared = new ReactiveVar(null)
+  @preparing = new ReactiveVar(false)
+
+  @autorun =>
+    if @prepared.get()
+      @preparing.set(false)
+
 Template.downloadCSVModal.helpers
   getField: (row, field)->
     row[field]
 
+  preparing: ->
+    Template.instance().preparing.get()
+
 Template.downloadCSVModal.events
   'click .download-csv': (event, instance) ->
+    instance.prepared.set(null)
+    instance.preparing.set(true)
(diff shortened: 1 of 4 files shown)


client/controllers/downloadCSVModal.coffee
```diff
@@ -1,10 +1,29 @@
...
 Template.downloadCSVModal.events
   'click .download-csv': (event, instance) ->
+    instance.prepared.set(null)
+    instance.preparing.set(true)
     fileType = $(event.currentTarget).attr('data-type')
     table = instance.$('table')
     if table.length
-      table.tableExport(type: fileType)
+      # Delay export so UI can respond to change in reactiveVars
+      setTimeout ->
```
jgoley: @nathanathan Is there a better way to handle this? I suppose the logic in the event runs immediately after the reactiveVar is set and blocks the reactivity (the helper and subsequently the UI change)?

nathanathan: A web worker might work for this but I don't think it's worth the trouble.

client/controllers/downloadCSVModal.coffee
```diff
@@ -1,10 +1,29 @@
+Template.downloadCSVModal.onCreated ->
+  @prepared = new ReactiveVar(null)
+  @preparing = new ReactiveVar(false)
```
nathanathan: The it looks like the preparing variable isn't necessary since it could be derived from the prepared variable.

jgoley: Good call! I suppose I was overthinking it. 2d6c2ae makes this change.

nathanathan:
👍

PR #391, opened 2017-06-27
Let users delete documents they added

nathanathan:
Pivotal story: https://www.pivotaltracker.com/story/show/143426399


jgoley:
Looks good 👍 
I made a small change—just moved the delete button to directly above the text. I thought it got a little lost on the far right.

PR #394, opened 2017-06-30
Replace bootstrap-validator with parsley.js

jgoley:
[PT Chore](https://www.pivotaltracker.com/story/show/147722207)


nathanathan:
The document's title field is being required now, but I think we made the decision that it should be optional. Also, the validation on the publication time isn't stopping submits.

jgoley:
Thanks for spotting those issues. Should be good to go now.

nathanathan:
👍 Looks good

PR #395, opened 2017-07-05
Fix incident modal content repositioning 147827315

jgoley:
This took a good bit of digging. Applies `will-change: transform` to scrollable elements within the `suggestedIncidentModal` which tells the browser to expect changes to these elements.


nathanathan:
👍

PR #397, opened 2017-07-07
Fix deleting empty documents

jgoley:
Adds a delete button within the little error notification and moves the button up to the top header.


nathanathan:
👍

PR #398, opened 2017-07-07
Fix bugs related to location not having name

jgoley:
There was an edge case where if a location did not have a name property, an error would be thrown and a phantom space would appear in the UI.
Also fixes an issue which disabled the incident detail toggle when the incident had events but no locations.

[PT Bug](https://www.pivotaltracker.com/story/show/148483127)


nathanathan:
👍

PR #399, opened 2017-07-10
Fix padding of suggested and parsley-error form elements

jgoley:
[PT Bug](https://www.pivotaltracker.com/story/show/148483257)

### DIFF
--- a/imports/stylesheets/forms/parsley.import.styl
+++ b/imports/stylesheets/forms/parsley.import.styl
@@ -28,8 +28,8 @@ $invalid = darken($danger, 10%)
     background lighten($danger, 95%)
     border-top 1px solid lighten($danger, 75%)
     border-color lighten($danger, 75%) !important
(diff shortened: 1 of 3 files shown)


imports/stylesheets/forms/parsley.import.styl
```diff
@@ -28,8 +28,8 @@ $invalid = darken($danger, 10%)
     background lighten($danger, 95%)
     border-top 1px solid lighten($danger, 75%)
     border-color lighten($danger, 75%) !important
-    padding-top 10px !important
-    padding-bottom 10px !important
+    padding-top .75em !important
```
nathanathan: Could we do away with the !important modifiers here?

jgoley: Yep, good call!

nathanathan:
👍

PR #401, opened 2017-07-10
Fix errors realted to 0 case/death counts

jgoley:
[PT Bug](https://www.pivotaltracker.com/story/show/148425141)


nathanathan:
Editing an incident and setting deaths to 0 doesn't work when cases was previously set.

jgoley:
b43da58 Should take care of the noted issue.

nathanathan:
👍

PR #402, opened 2017-07-10
Fix add new incident button

jgoley:
[PT Bug](https://www.pivotaltracker.com/story/show/148583391)

### DIFF
--- a/client/views/incidentReports/incidentForm.jade
+++ b/client/views/incidentReports/incidentForm.jade
@@ -6,17 +6,17 @@ template(name="incidentForm")
       if documentId
         input(type="hidden" name="articleId" value=documentId)
       unless confirmation
-        if documentId
+        if documentUrl
           .form-group.form-group--full
             +documentInputLabel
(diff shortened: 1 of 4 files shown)


client/views/incidentReports/incidentForm.jade
```diff
@@ -3,23 +3,24 @@ template(name="incidentForm")
     label.control-label.featured.container-flex.no-break
       h5 Incident Details
     .add-incident--wrapper
-      if documentId
-        input(type="hidden" name="articleId" value=documentId)
-      unless confirmation
+      unless documentFromTextContent
```
nathanathan: I think this will cause problems when editing incidents on documents from text since there will be no articleId input. It would be better to update the inputs and article selector to handle text based documents than create separate logic for them.

jgoley: Is e1e27eb an okay solution?

nathanathan: Yes

nathanathan:
👍

PR #405, opened 2017-07-11
Fix matching start dates creating zero length subintervals

nathanathan:
This bug lead to infinite rates that could break the EpiCurve plot like on this event: https://eidr-connect.eha.io/events/curated-events/gfnPs88SBb3aaBeeA


jgoley:
👍

PR #406, opened 2017-07-12
Remove save and duplicate event modal button

nathanathan:
Pivotal story: https://www.pivotaltracker.com/story/show/148468445

### DIFF
--- a/client/controllers/incidentReports/incidentModal.coffee
+++ b/client/controllers/incidentReports/incidentModal.coffee
@@ -45,7 +45,6 @@ Template.incidentModal.events
     # button's default is prevented
     $('#add-incident').submit()
     return if not instance.valid.get()
-    duplicate = $(event.target).hasClass('save-modal-duplicate')
     form = instance.$('form')[0]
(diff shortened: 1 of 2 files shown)


jgoley:
While testing I noticed that saving an incident on with a good many incidents there's a significant delay in saving the incident and dismissing the modal (http://localhost:3000/events/curated-events/gtT3A6ozJsvzHdzWp)
Looks good otherwise 👍

client/controllers/incidentReports/incidentModal.coffee
```diff
@@ -68,13 +67,12 @@ Template.incidentModal.events
...
           $('.reactive-table tr').removeClass('open')
           $('.reactive-table tr.tr-details').remove()
-          if !duplicate
-            form.reset()
-            notify('success', 'Incident added.')
-            if instance.data.offCanvas
-              stageModals(instance, instance.modals)
-            else
-              Modal.hide('incidentModal')
+          form.reset()
```
jgoley: I wonder if the reset is necessary considering the form is completely removed from the DOM after the modal is dismissed?

PR #407, opened 2017-07-12
Fix incident to text with 0 count in map and extracted incidents

jgoley:
[PT Bug](https://www.pivotaltracker.com/story/show/148748231)


nathanathan:
👍

PR #409, opened 2017-07-18
Handle GRITS API changes

nathanathan:
This makes some updates to the incident report suggestion code to handle some changes to the resolvedKeyword data returned by the GRITS API. This also adds species suggestions and fixes a couple bugs. The annotated content's end offset was off by one and a environment variable defined constant was being used on the client side where it is unavailable.


jgoley:
👍

PR #410, opened 2017-07-19
Event page filtration

jgoley:
Adds filter sidebar and filtration functionality for incident type, status and "other properties" (only travel related at this point). 

I'm not sure how I feel about having all the selectors unchecked by default. I might do a little research to determine if this is the best approach.

[PT Story](https://www.pivotaltracker.com/story/show/148836369)


nathanathan:
👍 Looks good!

PR #412, opened 2017-07-20
Event page date filtration

jgoley:
Builds on the event page filtration and adds a date range slider

[PT Story](https://www.pivotaltracker.com/story/show/148836407)


jgoley:
@nathanathan I'm not sure why the _Add incident report_ scenario is failing—it's passing for me locally using both chrome and phantomjs.

nathanathan:
My guess would be that the stylesheets for the slider are doing something unexpected on the server (perhaps due to a difference in screen size) or one of the helpers removed from curatedEvent.coffee was being used somewhere. Perhaps adding some logging around the failing test to print out error messages would help track down the issue.

tests/features/step_definitions/incidents_steps.coffee
```diff
@@ -23,7 +23,6 @@ do ->
       # Location
       @client.setValue('input.select2-search__field', 'f')
       @client.clickWhenVisible('.select2-results__option--highlighted')
-      @client.pause(40000)
```
nathanathan: I would try leaving in a pause here, although I don't think it should be 40 seconds.

jgoley: I'll try that. My bad with setting it to 40 seconds.

jgoley:
@nathanathan Adding back the pause did the trick!

nathanathan:
👍 Looks great!

PR #413, opened 2017-07-20
Update loading indicator of resolved plot - add veil

jgoley:
Small update to the resolved incidents plot that adds a veil and brings the loading animation to the front in space.

[PT Bug](https://www.pivotaltracker.com/story/show/149273195)


nathanathan:
👍

PR #414, opened 2017-07-21
Add choropleth layer to affected areas map

nathanathan:
Story: https://www.pivotaltracker.com/story/show/148836615


jgoley:
I'm not seeing the choropleth or marker layers.
Looks like `/world.geo.json` is required?

nathanathan:
@jgoley Sorry, I forgot to add the geo.json file. The marker layers aren't implemented this this PR.

jgoley:
Question: when the template loads, the geo.json file downloads. What happens if the user interacts with the map before it's finished?

Looks good! 👍

jgoley:
Also: We should probably zoom the map out to contain all the highlighted countries if the view does not initially contain them?

nathanathan:
@jgoley If the user interacts with the map before the geojson loads the choropleth layer will load when it's finished and their viewport will be preserved.

Zooming the map around when the layers change might not be what users want.

PR #416, opened 2017-07-24
Add location filtration to event page

jgoley:
[PT Story](https://www.pivotaltracker.com/story/show/148836467)

### DIFF
--- a/client/controllers/events/eventFiltration.coffee
+++ b/client/controllers/events/eventFiltration.coffee
@@ -1,10 +1,30 @@
+import EventIncidents from '/imports/collections/eventIncidents'
+
 Template.eventFiltration.onCreated ->
   @PROP_PREFIX = 'filter-'
   @typeProps = ['cases', 'deaths']
   @statusProps = ['suspected', 'confirmed', 'revoked']
+  @locationLevels = [
+    {
+      prop: 'countryName'
+      name: 'Country Name'
+    }
+    {
+      prop: 'admin1Name'
+      name: 'Admin Name 1'
+    }
+    {
+      prop: 'admin2Name'
+      name: 'Admin Name 2'
+    }
+  ]
   @types = new ReactiveVar([])
   @status = new ReactiveVar([])
   @properties = new ReactiveVar({})
+  @countryLevel = new ReactiveVar('countryName')
+  @selectedLocations = new Meteor.Collection(null)
+  @locations = new Meteor.Collection(null)
+
   @autorun =>
     filters = {}
     types = @types.get()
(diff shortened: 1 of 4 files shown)


client/controllers/events/eventFiltration.coffee
```diff
@@ -60,3 +119,24 @@ Template.eventFiltration.events
     instance.$('.other-properties input:checked').each (i, input) ->
       otherProps[instance.removePropPrefix(input.id)] = true
     instance.properties.set(otherProps)
+
+  'change .locations select': (event, instance) ->
+    instance.countryLevel.set(event.target.value)
+
+  'click .location-list li': (event, instance) ->
+    selectedLocations = instance.selectedLocations
+    if selectedLocations.findOne(id: @_id)
```
nathanathan: Could we store the id as "_id" instead so the query specifier isn't needed?

client/controllers/events/eventFiltration.coffee
```diff
@@ -23,6 +43,12 @@ Template.eventFiltration.onCreated ->
       filters.status =
         $in: status
 
+    selectedLocations = @selectedLocations.find().map (location) ->
+      location.name
+    if selectedLocations.length
+      locationProjection = "locations.#{@countryLevel.get()}"
+      filters[locationProjection] = $in: selectedLocations
```
nathanathan: It's an edge case, but It's possible for two separate administrative divisions to share an admin1name or admin2name. The selected location's parent admin name and country should be included in the filter to distinguish between locations that share the same name.

client/controllers/events/eventFiltration.coffee
jgoley: Yep! Seems like I had tried that in the past and ran into issues.

client/controllers/events/eventFiltration.coffee
jgoley: How does this look for a solution: f2179b4
I also noticed there's a case where a continent can be used. I suppose we should consider this?

client/controllers/events/eventFiltration.coffee
jgoley: I ran into an issue with duplicate ids (I think this is the issue I had in the past). Any way around it?

nathanathan: @jgoley That might be a sign that something is getting added to the collection twice or that the same id is being used by multiple locations. Doing a `if not selectedLocations.find(x)` before all your inserts should prevent any duplicates from being added. If it's something else using the id property is okay.

jgoley: I changed the insert into upsert: https://github.com/ecohealthalliance/eidr-connect/pull/416/commits/10d195883d1470114afe63d1c701eab724438452#diff-91529dcc4459ea5068521029352eb3b6R69

client/controllers/events/eventFiltration.coffee
nathanathan: @jgoley The geonames don't have information we can use to determine containment for continent and region names like "West Africa," so I think it will be difficult to do. I'll put something in the backlog for it.

nathanathan:
👍 Looks good

PR #417, opened 2017-07-27
Refactor location select2

nathanathan:
This makes it so the locationSelect2 template uses the generic select2 template. It also fixes a bug where geoname lookup service urls passed in as environment vars wouldn't be used and it fixes some bugs that caused the "values" reactive vars used in select2 templates to have incorrect values.

### DIFF
--- a/client/controllers/incidentReports/incidentForm.coffee
+++ b/client/controllers/incidentReports/incidentForm.coffee
@@ -1,6 +1,7 @@
 import Articles from '/imports/collections/articles.coffee'
 import createInlineDateRangePicker from '/imports/ui/inlineDateRangePicker.coffee'
 import {
+  formatLocation,
   keyboardSelect,
   removeSuggestedProperties,
   diseaseOptionsFn } from '/imports/utils'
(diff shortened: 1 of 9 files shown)


client/controllers/incidentReports/incidentForm.coffee
```diff
@@ -59,6 +60,12 @@ Template.incidentForm.onCreated ->
 
     @incidentStatus.set(@incidentData.status or '')
 
+    @locations.set(@incidentData.locations.map (loc)->
+      id: loc.id
+      text: formatLocation(loc)
+      item: loc
+    )
```
jgoley: Just suggesting some minor code style updates: remove the surrounding parens and add a space after the arguments?

client/controllers/select2.coffee
```diff
@@ -45,15 +45,28 @@ Template.select2.onRendered ->
...
+    selectedValues = instance.$("select").select2('data')
+    uniqueValues = _.uniq(_.pluck(instance.values.get(), 'id'))
+    uniqueSelectedValues = _.uniq(_.pluck(selectedValues, 'id'))
+    intersection = _.intersection(uniqueValues, uniqueSelectedValues)
+    if intersection.length != uniqueSelectedValues.length or uniqueValues.length != uniqueSelectedValues.length
+      instance.values.set(selectedValues.map (data)->
+        id: data.id
+        text: data.text
+        item: data.item
+      )
```
jgoley: Code style updates here also?

client/controllers/incidentReports/incidentForm.coffee
nathanathan: I think the they make this a little more readable. In general, I don't think we should worry about whether parenthesis are used.

jgoley: 👍

PR #418, opened 2017-07-27
Add incident modified date, fix unmarked cumulative counts

nathanathan:
This marks incidents from pre December 2016 events as cumulative if they report more than 50 cases in a day.

This also prevents validation from running during a database update.


jgoley:
Good call with adding modified date to incidents.
👍  Looks good!

PR #419, opened 2017-07-28
Repair smart event pages

nathanathan:
This makes it so the updates to the curated event pages are available on the smart event pages. It also adds separate disassociate/delete incident controls to the event pages.


client/controllers/events/eventIncidentReports.coffee
```diff
@@ -219,32 +218,52 @@ Template.eventIncidentReports.events
...
+        @$('.tooltip').remove()
+        notify('success', 'Incidents Deleted')
+        Modal.hide('confirmationModal')
+    Modal.show 'confirmationModal',
+      primaryMessage: 'Are you sure you want to completely remove this incident?'
+      secondaryMessage: """
+        Deleting an incident removes it from all events in EIDR-Connect.
+        This action should only be taken if the incident contains incorrect data.
+        You can instead remove it from this event alone by unlinking it.
+        Are you sure you want to delete it?
```
jgoley: This may be a little confusing while on a smart event page because the user can't really unlink an incident without modifying the event details.

PR #420, opened 2017-07-31
Only sync curated event data

nathanathan:
This updates the one way sync code to only sync data from curated events. The database is becoming too large to sync all the data.


jgoley:
👍  Looks good!

PR #421, opened 2017-07-31
Fix add new incident - check if incident has locations

jgoley:
Fixes [this bug](https://www.pivotaltracker.com/story/show/148583391) — Checks if incident has locations before mapping.


nathanathan:
👍

PR #422, opened 2017-07-31
Add content to event references tab

nathanathan:
Story: https://www.pivotaltracker.com/story/show/149137839


jgoley:
👍 
I made some small changes to the interactivity of the table.
I guess at some point we should connect event documents to the documents page (curator inbox)?

PR #423, opened 2017-07-31
Align dates in filtration with dates in incident

jgoley:
@nathanathan I suppose the dates in the table are correct? (those formatted in formatDateRange). So I also use UTC to display the dates in the filtration area.


nathanathan:
👍 Making the locale for the filter dates and table dates consistent is a definite improvement. Maybe we should change all the dates to use the browser's locale in the future.

PR #425, opened 2017-08-01
Event filtration no results

jgoley:
Makes improvements to UX by adding no results messages and loading indicators to let user know filters have been considered in the results.

### DIFF
--- a/client/controllers/events/eventIncidentReports.coffee
+++ b/client/controllers/events/eventIncidentReports.coffee
@@ -22,6 +22,7 @@ Template.eventIncidentReports.onCreated ->
   @subscribe('feeds')
   @plotZoomed = new ReactiveVar(false)
   @dataLoading = new ReactiveVar(false)
+  @rendering = new ReactiveVar(false)
   # underscore template for the mouseover event of a group
   @tooltipTmpl = """
     <% if ('applyFilters' in obj) { %>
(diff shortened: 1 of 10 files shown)


client/views/events/curatedEvent.jade
```diff
@@ -1,20 +1,23 @@
...
+        h2 This event has been deleted
+    else
+      section.event--content-wrapper.container-flex.no-break
+        aside.event--filtration
+          +eventFiltration(
+            filterQuery=filterQuery
+            selectedIncidentTypes=selectedIncidentTypes)
+        .event--content-container
+          unless results
+            .veil.modal-layer
```
nathanathan: The veil is covering the eventNav

client/controllers/events/eventIncidentReports.coffee
```diff
@@ -48,6 +49,17 @@ Template.eventIncidentReports.onCreated ->
     <% } %>
   """
 
+  @autorun =>
+    # When user modifies query, set rendering to true which triggers a loading
```
nathanathan: It might simplify things to handle the feedback for filter changes in the filter or event template.

client/views/events/curatedEvent.jade
jgoley: Oh...I forgot about the "details" and "references" nav items.

client/controllers/events/eventIncidentReports.coffee
jgoley: I thought about going that route but the incidents view is the only view without a loading indicator that appears when the filtration criteria changes. So I figured we should add one for consistency.

nathanathan:
👍 Looks good

PR #426, opened 2017-08-01
Update affected areas marker color to increase contrast

jgoley:
[PT Story](https://www.pivotaltracker.com/story/show/149695433)


nathanathan:
👍

PR #427, opened 2017-08-01
Restrict height of event filter locationn list

jgoley:
[PT Bug](https://www.pivotaltracker.com/story/show/149846442)


nathanathan:
👍

PR #428, opened 2017-08-01
Always show slider to keep UI consistent

jgoley:
[PT Story](https://www.pivotaltracker.com/story/show/149672783)


nathanathan:
👍

PR #429, opened 2017-08-02
Extend geoname containment to use muti-country regions

nathanathan:
This extends the logic for incident count resolution and location filter queries so that geonames for continents can be used.

Story: https://www.pivotaltracker.com/story/show/149547697


client/controllers/events/eventFiltration.coffee
```diff
@@ -76,6 +81,11 @@ Template.eventFiltration.onCreated ->
     countryLevel = @countryLevel.get()
     selectedLocations = @selectedLocations.find().map (location) ->
       query = {}
+      if countryLevel == 'region'
+        regionInfo = _.findWhere(regionToCountries, name: location.region)
+        console.log location, regionToCountries
```
jgoley: Remove this log?

jgoley:
👍  Looks good other than that log.

PR #430, opened 2017-08-02
Event page tests

jgoley:
Adds a test scenario which tests each filtration section. Thoughts on adding a scenario which combines several filter properties?
Also adds:
- A clear button to clear all the filters
- Made the incident list reactive so when a new incident is added, it will reactively appear in the list
- Only add the location query if default range and selected not equal

[PT Chore](https://www.pivotaltracker.com/story/show/149213109)

### DIFF
--- a/client/controllers/events/eventFiltration.coffee
+++ b/client/controllers/events/eventFiltration.coffee
@@ -114,7 +114,7 @@ Template.eventFiltration.onRendered ->
     endDate = new Date(selectedRange[1])
     isDefault = moment(defaultRange[0]).isSame(startDate) and
       moment(defaultRange[1]).isSame(endDate)
-    if not isDefault
+    if not isDefault and EventIncidents.find().count() > 1
       filters['dateRange.start'] =
         $lte: endDate
       filters['dateRange.end'] =
--- a/client/controllers/events/eventFiltration.coffee
+++ b/client/controllers/events/eventFiltration.coffee
@@ -5,7 +5,7 @@ formatDateForInput = (date) ->
   unless date
     return moment()
   date = if date.getTime then date else new Date(Math.ceil(date))
-  moment(date)
+  moment.utc(date)
 
 Template.eventFiltration.onCreated ->
   @PROP_PREFIX = 'filter-'
@@ -62,10 +62,11 @@ Template.eventFiltration.onRendered ->
     defaultRange = @eventDateRange.get()
     ['start', 'end'].forEach (input, i) =>
       $input = @$(".#{input}-date").data("DateTimePicker")
-      if defaultRange and (defaultRange[0] instanceof Date and defaultRange[1] instanceof Date)
-        $input.minDate(moment(defaultRange[0]).subtract(1, 'day'))
-        $input.maxDate(moment(defaultRange[1]).add(1, 'day'))
+      if defaultRange and not _.isNumber(defaultRange[0])
+        $input.minDate(moment(defaultRange[0]))
+        $input.maxDate(moment(defaultRange[1]))
       $input.date(formatDateForInput(selectedRange[i]))
+
   # Establish and update date ranges when incidents collection changes
   @autorun =>
     incidents = EventIncidents.find({}, fields: 'dateRange': 1).fetch()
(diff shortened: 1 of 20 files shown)


packages/_tests/fixtures.coffee
```diff
@@ -108,3 +132,24 @@ if Meteor.isAppTest
       catch error
         # this user shouldn't belong in the production database
         console.warn("TestingAdmin user '#{email}' exists")
+
+    addIncidents: (eventId, incidentCount) ->
+      Meteor.call 'addEventSource', testSource2, eventId, (error, articleId) ->
```
nathanathan: Since these methods run on the server they can be called synchronously without a callback. Using async calls might cause timing issues in the tests since the addIncidents method will finish before the async calls complete.

jgoley:
@nathanathan Do you mind taking a look at this—I'm running into a few issues.
I went ahead and switched the date selector to the bootstrap version and in doing so am running into trouble with reactively setting the value of the input dates after a new incident is added.

jgoley:
The last commit fixes an issue which caused an infinite loop when the event page loaded (The dates being compared in the datepicker change event never matched up, i think because one was flagged as locale which caused a discrepancy?)
There are a few issues: 
For some reason the dates are updated just by clicking on the start date input. (i suppose this is mostly a UX issue)
During the tests, the datepicker seems to break—for some reason `selectedDateRange`  is not updated when the date input value changes. (even though it works fine otherwise)

client/controllers/slider.coffee
```diff
@@ -10,44 +10,47 @@ Template.slider.onRendered ->
...
+      connect: true
       range:
-        min: newMaxMin[0]
-        max: newMaxMin[1]
+        min: formattedSliderRange[0]
+        max: formattedSliderRange[1]
+
+    sliderEl.noUiSlider.on 'change', _.debounce (values, handle) =>
+      @data.selectedRange.set values
+      # Show or hide the left/right slider icon
```
nathanathan: This is already handled by reactivity in the render function. You don't need any of the jquery code here.

client/controllers/events/eventFiltration.coffee
```diff
@@ -228,8 +241,24 @@ Template.eventFiltration.events
   'click .locations .deselect-all': (event, instance) ->
     instance.selectedLocations.remove({})
 
-  'change .dates input': (event, instance) ->
-    dates = []
-    instance.$('.dates input').each (i, input) ->
-      dates.push(new Date(input.value))
-    instance.selectedDateRange.set(dates)
+  'dp.change': (event, instance) ->
+    startStr = $('.start-date').data('DateTimePicker').date()?.utc().format('YYYY-MM-DD')
```
nathanathan: The date time picker displays locale dates. Interpreting them as utc could cause them to have different values that what the user entered.

client/controllers/events/eventFiltration.coffee
```diff
@@ -3,9 +3,9 @@ import regionToCountries from '/imports/regionToCountries.json'
 
 formatDateForInput = (date) ->
   unless date
-    return moment.utc().format('YYYY-MM-DD')
+    return moment()
```
nathanathan: Why was this function modified?

nathanathan:
selectedDateRange might not be updating during the tests because the date input never looses focus and triggers a change event. If that's not the issue making the tests use the calendar widget instead of entering text directly into the date field might work.

client/controllers/events/eventFiltration.coffee
jgoley: I was trying to figure out why there is a discrepancy when the dates are compared in the picker event.

jgoley: Looks like the start date is consistently a day off.

jgoley:
@nathanathan I switched to the daterangepicker for a slightly better UX and because it allows us a little more control in managing reactivity.
I also added a comparison of the selected date range with the event's default date range in the filter query autorun so we can keep the filterQuery reactiveVar empty when filters are not being used.
I also reverted back to updating the slider's max and min when the range changes rather than creating a new instance of the slider. When it is updated, the slider handle's movement is animated which I think is a better user experience.

nathanathan:
👍 Looks good

PR #431, opened 2017-08-03
Misc. bug fixes

nathanathan:
This resolves the document title issue [here](https://www.pivotaltracker.com/story/show/148683711) and a few other issues described in the commit messages.

### DIFF
--- a/client/controllers/events/eventIncidentReports.coffee
+++ b/client/controllers/events/eventIncidentReports.coffee
@@ -53,7 +53,7 @@ Template.eventIncidentReports.onCreated ->
     # indicator to appear indicating filters have been considered in results.
     data = @data
     data.filterQuery.get()
-    if data.loaded.get()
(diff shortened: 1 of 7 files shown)


client/controllers/events/eventIncidentReports.coffee
```diff
@@ -53,7 +53,7 @@ Template.eventIncidentReports.onCreated ->
     # indicator to appear indicating filters have been considered in results.
     data = @data
     data.filterQuery.get()
-    if data.loaded.get()
+    if data.loaded and data.loaded.get()
```
jgoley: For future reference should we use this pattern or the existential operator? I suppose this compiles down to less code.

nathanathan: Either way would be fine. I didn't think to use the existential operator here.

PR #432, opened 2017-08-04
Remove invalid dates

nathanathan:
This removes invalid date ranges where the end occurs before the start and prevents them from being created. It also changes the measure of date precision to the length of the date interval.

### DIFF
--- a/server/updaters.coffee
+++ b/server/updaters.coffee
@@ -11,7 +11,7 @@ import feedSchema from '/imports/schemas/feed'
 import Constants from '/imports/constants.coffee'
 import { regexEscape } from '/imports/utils'
 
-DATA_VERSION = 9
+DATA_VERSION = 11
 AppMetadata = new Meteor.Collection('appMetadata')
 priorDataVersion = AppMetadata.findOne(property: "dataVersion")?.value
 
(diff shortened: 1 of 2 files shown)


server/updaters.coffee
```diff
@@ -191,5 +191,12 @@ module.exports = ->
         incident.modifiedDate = new Date()
         Incidents.update(_id: incident._id, incident)
 
+  console.log "Removing incidents with invalid dates..."
+  invalidDateIncidents = []
+  Incidents.find().map (incident) ->
+    if incident.dateRange.end < incident.dateRange.start
+      invalidDateIncidents.push(incident._id)
+  console.log "Removing #{invalidDateIncidents.length} incidents"
+  Incidents.remove(_id: $in: invalidDateIncidents)
```
jgoley: Should we remove the entire incident? What if the incident's other data is accurate?

nathanathan: The incidents are kind of all or none when it comes to the location, count and date info. If they're missing one those those fields we can't use them for anything, so I'm alright with removing the full incident. Also, these are mostly or all auto-generated so we'll could eventually regenerate them after the processing code has been updated.

jgoley: I see. Thanks for the explanation!

jgoley:
👍

PR #433, opened 2017-08-07
Fix species name bug and location filter bug

nathanathan:
Prior to this update [Object] was being displayed in the species column of the extract-incidents table and the location filter was not showing all the available locations when initially loaded.


jgoley:
👍

PR #436, opened 2017-08-09
Fix incident details row removal

nathanathan:
Story: https://www.pivotaltracker.com/story/show/149903339

### DIFF
--- a/client/controllers/events/eventIncidentReports.coffee
+++ b/client/controllers/events/eventIncidentReports.coffee
@@ -139,6 +139,10 @@ Template.eventIncidentReports.onRendered ->
       @updatePlot(groups)
       return
 
+  @autorun =>
+    EventIncidents.find(Template.instance().data.filterQuery.get())
+    $('tr.details').remove()
+
 Template.eventIncidentReports.helpers
   getSettings: ->
     tableName = 'event-incidents'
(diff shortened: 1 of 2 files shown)


client/controllers/events/eventIncidentReports.coffee
```diff
@@ -139,6 +139,10 @@ Template.eventIncidentReports.onRendered ->
       @updatePlot(groups)
       return
 
+  @autorun =>
+    EventIncidents.find(Template.instance().data.filterQuery.get())
+    $('tr.details').remove()
```
jgoley: Could we move this up to the `autorun` on line 122 or should we keep it separate?

client/controllers/events/eventResolvedIncidents.coffee
```diff
@@ -149,17 +150,30 @@ Template.eventResolvedIncidents.onRendered ->
...
+    differentialIncidents = convertAllIncidentsToDifferentials(allIncidents)
+    @differentialIncidents.set(differentialIncidents)
+
+  @autorun =>
     @hoveredIntervalClickEvent = null
     incidentType = @incidentType.get()
-
+    differentialIncidents = @differentialIncidents.get()
+    # Toggle the incident type if there are no matching incidents
+    console.log differentialIncidents
```
jgoley: Remove?

client/controllers/events/eventIncidentReports.coffee
nathanathan: I think it should be separate. Although the dependencies are currently the same that might not always be the case because the autoruns control separate components.

PR #437, opened 2017-08-09
Add veil to filters if no incidents

jgoley:
[PT Bug](https://www.pivotaltracker.com/story/show/149962130)

### DIFF
--- a/client/controllers/smartEvents/smartEvent.coffee
+++ b/client/controllers/smartEvents/smartEvent.coffee
@@ -1,4 +1,5 @@
 import SmartEvents from '/imports/collections/smartEvents'
+import EventIncidents from '/imports/collections/eventIncidents'
 #Allow multiple modals or the suggested locations list won't show after the
 #loading modal is hidden
 Modal.allowMultiple = true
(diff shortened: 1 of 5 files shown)


nathanathan:
This makes it so there is no way to change the filters if a filter is selected that removes all the incidents.

jgoley:
My bad. The latest commit fixes that by using the `hasNoIncidents` method on the instance. It also adds the same logic to the smart event page.

client/controllers/smartEvents/smartEvent.coffee
```diff
@@ -11,6 +12,10 @@ Template.smartEvent.onCreated ->
   @filterQuery = new ReactiveVar({})
   @selectedIncidentTypes = new ReactiveVar([])
 
+  @hasNoIncidents = =>
+    filterQuery = @filterQuery.get()
+    _.isEmpty(filterQuery) and not EventIncidents.find(filterQuery).count()
```
nathanathan: Does the filterQuery being non-empty imply that the event has incidents?

jgoley: Oh, right. I guess we only need to check if the empty query has a count.
This brings up an issue—if the query is empty (the user creates a smart event with only a title) doesn't the smart event page try to subscribe to every incident?

nathanathan:
👍

PR #438, opened 2017-08-10
Switch to scatter plot popups that can open incidents

nathanathan:
Story: https://www.pivotaltracker.com/story/show/149845451


jgoley:
Noticed a few things:
- When viewing the incidents then you switch tabs then go back to incidents, when opening incident details this error occurs: `Uncaught Error: Must be attached`

- When the template is loaded, popup is visible: 
screen shot 2017-08-10 at 11 13 55 am

- The sizing and the position of the popup when near the edge of the viewport or container is pretty skinny. Before we positioned it to the left in a case like this.
screen shot 2017-08-10 at 11 20 35 am

jgoley:
I made some small changes to the tooltip and added scaling on mouseover/out.
Looks good 👍

PR #439, opened 2017-08-15
Disassociate deleted incidents from events

jgoley:
[Related PT Bug](https://www.pivotaltracker.com/story/show/150228709)


nathanathan:
👍

PR #440, opened 2017-08-15
Add incident to incident marker area if not in array

jgoley:
Not sure if this is necessary?...it only seems to be a problem on [this event](https://eidr-connect.eha.io/events/curated-events/gfnPs88SBb3aaBeeA).

### DIFF
--- a/client/controllers/events/eventAffectedAreas.coffee
+++ b/client/controllers/events/eventAffectedAreas.coffee
@@ -154,6 +154,8 @@ Template.eventAffectedAreas.onRendered ->
       incident.locations.forEach (location) ->
         key = "#{location.latitude},#{location.longitude}"
         incidentsByLatLng[key] = (incidentsByLatLng[key] or []).concat(incident)
+    for key, incidents of incidentsByLatLng
+      incidentsByLatLng[key] = _.uniq(incidents, false, (x) -> x._id)
     if @markerLayer.get()
       markerLayerGroup = L.layerGroup()
       for key, incidents of incidentsByLatLng


client/controllers/events/eventAffectedAreas.coffee
```diff
@@ -153,7 +153,8 @@ Template.eventAffectedAreas.onRendered ->
     EventIncidents.find(@data.filterQuery.get()).map (incident) ->
       incident.locations.forEach (location) ->
         key = "#{location.latitude},#{location.longitude}"
-        incidentsByLatLng[key] = (incidentsByLatLng[key] or []).concat(incident)
+        if not incidentsByLatLng[key]
```
nathanathan: This makes it so only one incident can be at a coordinate.

jgoley: I see. Should we ignore the duplicate id warning?

nathanathan: I'm not seeing the duplicate id warning. Could you give an example of how to reproduce it?

jgoley: This event is the only that I've found that has this issue (I guess the only event that has incidents which overlap): https://eidr-connect.eha.io/events/curated-events/gfnPs88SBb3aaBeeA/affected-areas
On the "Affected Areas" tab click incidents to reveal the markers. The warnings should appear in the console.

nathanathan: @jgoley The problem was that the same incident could be added to the same coordinate twice if it has two locations with the same coordinate. I added a commit that fixes this by deduplicating the incidents for each coordinate.

PR #443, opened 2017-08-15
Do not allow users to delete default feed

jgoley:
[PT Bug](https://www.pivotaltracker.com/story/show/144114949)


nathanathan:
👍

PR #444, opened 2017-08-16
Update incident type selector, add active cases type

nathanathan:
Story: https://www.pivotaltracker.com/story/show/150013165

This also adds a type property to incidents which should be useful for reducing the amount of type inference code in the future.

### DIFF
--- a/client/controllers/incidentReports/incidentForm.coffee
+++ b/client/controllers/incidentReports/incidentForm.coffee
@@ -28,7 +28,8 @@ Template.incidentForm.onCreated ->
   if articleId
     @subscribe 'incidentArticle', articleId
   @incidentStatus = new ReactiveVar('')
-  @incidentType = new ReactiveVar('')
+  @incidentType = new ReactiveVar('caseCount')
+  @dateRangeType = new ReactiveVar('day')
   @locations = new ReactiveVar([])
   @suggestedFields = incident?.suggestedFields or new ReactiveVar([])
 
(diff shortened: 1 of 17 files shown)


client/controllers/incidentReports/incidentForm.coffee
```diff
@@ -82,43 +90,62 @@ Template.incidentForm.onRendered ->
...
-    Template.instance().incidentType.get() is 'other'
+    Template.instance().incidentType.get() is 'specify'
+
+  showRangeTab: ->
+    type = Template.instance().incidentType.get()
+    type in [
+      'caseCount'
+      'deathCount'
+      'specify'
+    ]
```
jgoley: Could we avoid declaring the `type` variable since its only referenced once? 
Also on line 105.

client/controllers/select2.coffee
```diff
@@ -1,5 +1,6 @@
 Template.select2.onCreated ->
   defaultOptionFn = (params, callback) =>
+    console.log @data
```
jgoley: Remove?

server/updaters.coffee
```diff
@@ -11,7 +11,7 @@ import feedSchema from '/imports/schemas/feed'
 import Constants from '/imports/constants.coffee'
 import { regexEscape } from '/imports/utils'
 
-DATA_VERSION = 13
+DATA_VERSION = 17
```
jgoley: I suppose it doesn't matter too much but should we drop this back down to 14?

client/controllers/incidentReports/incidentReport.coffee
```diff
@@ -15,11 +15,13 @@ Template.incidentReport.helpers
...
-
-  casesLabel: ->
-    pluralize 'Case', @cases, false
+  countLabel: ->
+    if @type == 'activeCount'
+      pluralize 'Active Case', @cases, false
+    else if @cases
+      pluralize 'Case', @cases, false
+    else if @deaths
+      pluralize 'Death', @deaths, false
```
jgoley: I think we should wrap the arguments in parens. I think the style guide recommends doing so if they can all fit on one line.

PR #447, opened 2017-08-18
Fix autogenerated incident updates

nathanathan:
This makes it so that if a document is reprocessed the new autogenerated incidents will be correctly merged with the old ones. Prior incidents that are not associated with an event and have not been modified by an user will be removed. New incidents with counts that overlap the remaining prior incidents will not be added.


frosario:
Going to close and reopen to retrigger Jenkins testing

server/methods.coffee
```diff
@@ -109,6 +154,9 @@ Meteor.methods
           # in the geoname lookup service's database.
           return
         geonamesById[geoname.id] = GeonameSchema.clean(geoname)
+        # Do not store alternate names because they aren't needed and take up
+        # a lot of space.
+        geoname.alternateNames = []
```
jgoley: This is a good idea. Should we remove the alternate names from existing incidents in a future chore?

imports/utils.coffee
```diff
@@ -288,6 +288,7 @@ export createIncidentReportsFromEnhancements = (enhancements, options) ->
         timeAnnotation.timeRange.end
       ).endOf('day')
       if timeAnnotation.beginMoment > timeAnnotation.endMoment
+        console.log(timeAnnotation)
```
jgoley: Remove?

client/router.coffee
```diff
@@ -48,7 +44,7 @@ Router.route "/admins",
   onBeforeAction: ->
     redirectIfNotAuthorized(@, ['admin'])
   waitOn: ->
-    Meteor.subscribe "allUsers"
+    [Meteor.subscribe("allUsers"), Meteor.subscribe("roles")]
```
jgoley: Just a comment: I wonder if we should consider moving this to the controller in the future?

nathanathan: The redirectIfNotAuthorized function in the router depends on this subscription so moving it might require more refactoring than you would expect.

imports/utils.coffee
nathanathan: I want to keep this log since it helps to track down the cause of "End date occurs before start date." errors.

server/methods.coffee
nathanathan: Yeah, that sounds good.

jgoley: https://www.pivotaltracker.com/story/show/150526265

imports/utils.coffee
jgoley: 👍

client/router.coffee
jgoley: Ohh, good point. I was looking to make things consistent but probably not worth the effort.
Let's leave it.

PR #449, opened 2017-08-29
Remove alternateLocations from incident locations

jgoley:
[PT Chore](https://www.pivotaltracker.com/story/show/150526265)


nathanathan:
👍

PR #452, opened 2017-09-13
Fix z-index issue with incident form tooltips

jgoley:
[PT Bug](https://www.pivotaltracker.com/story/show/151052811)


nathanathan:
👍

PR #453, opened 2017-09-13
Position incident details loading indicator relative to parent

jgoley:
[PT Bug](https://www.pivotaltracker.com/story/show/150521003)


nathanathan:
👍

PR #457, opened 2017-09-21
Add species filter to Event pages and smart events

nathanathan:
Pivotal story: https://www.pivotaltracker.com/story/show/150920423

This also places a limit on the number of incidents that can be published, and removes the "Select All" locations button from the filter panel because the "Deselect All" button has an equivalent effect on the filter.


arnoldbm:
+1

PR #458, opened 2017-09-25
Event page filter negation

jgoley:
Here's an initial attempt at establishing the UX and logic for 3 state filtering on the events page.
We probably need to add a simple explanation somehow—I'll do that tomorrow.

[PT story](https://www.pivotaltracker.com/story/show/150920530)


client/controllers/events/eventFiltration.coffee
```diff
@@ -6,6 +6,42 @@ formatDateForInput = (date) ->
     return moment()
   if date.getTime then date else new moment(date)
 
+# Update input classes based on state of button: unselected, positive, negative
+updateInputState = (event) ->
```
nathanathan: It seems like this should be handled by the jade template.

jgoley: Good point. I've been working on a refactor that reduces storing the state of the buttons in the DOM. It takes care of this issue but currently running into an issue with conflicting events. I'll push an update ASAP.

jgoley:
@nathanathan I think this is ready for final review.

nathanathan:
👍

PR #460, opened 2017-11-30
Add incident resolver option to convert regions into countries

nathanathan:
Aggregating cases by multiple country regions is undesirable when rendering maps. It may unnecessarily hide hotspots at the country level, regions may overlap creating inconsistencies in the overlapping area, and our contour data is for countries. To avoid these issues, this adds an option to the incident resolution code to replace incidents locations that reference a region with all the countries contained in that region. Note that our region data is for the most part limited to continents.

### DIFF
--- a/.config/chimp.js
+++ b/.config/chimp.js
@@ -27,12 +27,12 @@ module.exports = {
   screenshotsOnError: true,
   screenshotsPath: '.screenshots',
   captureAllStepScreenshots: false,
-  saveScreenshotsToDisk: false,
+  saveScreenshotsToDisk: true,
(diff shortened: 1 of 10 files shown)


toph-allen:
I can't really comment on certain parts of the commit, such as the changes to `server/router.coffee`.

I also can't comment on the change to the resolution code without more of a discussion, which I'll conduct in Slack.

toph-allen:
After discussion with Nathan, I get this a little more, and am gonna approve and merge it.

toph-allen:
Scratch that — Nathan is going to look into the test failure.

